Issue #3, opened 2026-07-28
CHANGELOG: `[0.6.0]` is still marked "unreleased" below seven shipped versions


ch-bas:
### What

In `CHANGELOG.md`, the oldest entry is:

```
## [0.6.0] — unreleased

Initial public distribution.
```

But it sits *below* seven later entries that are all dated and released — `0.7.0` (2026-07-20) through `0.10.0` (2026-07-25). A version labelled "unreleased" whose body is "Initial public distribution" can't be accurate once every version built on top of it has shipped; `0.6.0` was evidently the first public release and just never got its date filled in.

There's also a small convention mismatch: the file otherwise follows [Keep a Changelog](https://keepachangelog.com/), where an `## [Unreleased]` section lives at the **top** for upcoming changes — not attached to the oldest version at the bottom.

### Why it's worth fixing

The CHANGELOG is the customer-facing release history (its own header says so). A first-time reader scanning from the bottom sees the initial release marked "unreleased," which reads as an error and undercuts confidence in the rest of the history.

### Suggested fix

Give `0.6.0` its actual release date (whatever the first public tag's date was):

```diff
-## [0.6.0] — unreleased
+## [0.6.0] — 2026-07-XX

 Initial public distribution.
```

### Possibly related (worth a quick look, not asserting)

The latest git tag appears to be `v0.10.1`, while `package.json` (`0.10.0`), the README version badge (`0.10.0`), and the top CHANGELOG entry (`0.10.0`) all read `0.10.0`. If `0.10.1` shipped, it may be missing a CHANGELOG entry and a `package.json` version bump. Happy to fold that into the same PR if it's real.

I can open a small docs PR for the `0.6.0` date fix if that's useful.

PR #5, opened 2026-07-28
docs: sync the distribution to the 0.10.1 release

ch-bas:
## What changes

Syncs the distribution to the published v0.10.1 release (part of #3): adds the missing `[0.10.1] — 2026-07-27` CHANGELOG entry (adapted from the release notes) and bumps `0.10.0 → 0.10.1` in all nine version locations (`package.json`, the plugin manifests, the README badge, `setup.md`) so they agree with the published release, per the manifests-in-step rule in CONTRIBUTING.

The `[0.6.0] — unreleased` line itself is deliberately **not** touched: its first-ship date is not recoverable from this repository (no pre-0.10.1 tag, release, or event survives), so filling it in needs a maintainer who knows the date — that part stays open on #3.

## How it was verified

Docs/manifest-only change — no commands or camera behaviour affected. Verified with `grep -rn "0\.10\."` that every non-historical version reference now reads `0.10.1` and matches the latest published release (`gh release list` → `v0.10.1`, 2026-07-27). The CHANGELOG entry content is taken from the v0.10.1 GitHub release notes, not written from memory.

## Checklist

- [ ] `cargo test --workspace` passes — N/A: this repo contains no Rust source; docs/manifests only

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
     {
       "name": "reolink-cli",
       "description": "Operate Reolink cameras locally via reolink-cli: discover, login, inspect, configure, preview, snapshot, and control PTZ/light/audio/detection/recording.",
-      "version": "0.10.0",
+      "version": "0.10.1",
       "source": "./",
       "author": {
         "name": "Reolink"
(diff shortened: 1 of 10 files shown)


reolink-net (approved):
Verified against the live repo before approving.

#5's claim checks out: main still carried 0.10.0 in all nine version locations while v0.10.1 was the published release, and the CHANGELOG had no 0.10.1 entry. The added entry matches the release notes exactly — no invented content.

Thanks for catching this — the drift was mine.

PR #6, opened 2026-07-28
docs: drop the stale 'while this repository is private' install notes

### LINKED ISSUE #4: Docs: remove the "while this repository is private" install notes now that the repo is public
### What

The repository is now public (`visibility: public`), but a few places still tell readers the installers return 404 for anonymous callers and that a `GITHUB_TOKEN` is required "until it is made public." That guidance is stale and a little confusing for first-time setup — someone following the README's own one-line `curl … | sh` will read the note just below it and conclude the documented install won't work for them.

### Where

- **`README.md`**, "Installers" section — the blockquote:
  > **While this repository is private**, the one-line installers and the release download URLs return 404 for anonymous callers … Until it is made public, export a `GITHUB_TOKEN` …
- **`install.sh`**, header comment (~lines 11–12):
  ```sh
  # (While the repo is private, GitHub won't serve release assets over the public
  #  download URL — test with `gh release download` until it is public.)
  ```

(The generic fallback error strings — "set `GITHUB_TOKEN` if it is private" in `install.sh` / `install.ps1` — are still fine, since they only fire when the API can't resolve a release. No change needed there.)

### Suggested fix

Drop the two stale notes. If it's worth keeping a pointer for anyone who *does* hit a rate limit on the API, a one-liner is enough:

```diff
 ## Installers

-> **While this repository is private**, the one-line installers and the release
-> download URLs return 404 for anonymous callers — GitHub does not serve private
-> content that way. Until it is made public, export a `GITHUB_TOKEN` with access
-> to this repository before running an installer, or fetch the archive with
-> `gh release download`.
-
-
 `install.sh` and `install.ps1` fetch and run executables, so here is exactly
 what they do:
```

```diff
 # Env overrides: REOLINK_REPO=owner/repo, REOLINK_PREFIX=/install/root.
-# (While the repo is private, GitHub won't serve release assets over the public
-#  download URL — test with `gh release download` until it is public.)
 set -eu
```

Happy to open a small PR for this if useful.

ch-bas:
## What changes

Fixes #4. The repository is public now, but the README's Installers section and the `install.sh` header comment still said the installers return 404 for anonymous callers and that a `GITHUB_TOKEN` is needed "until it is made public". Both stale notes are removed.

The conditional fallback error strings ("set `GITHUB_TOKEN` if it is private") in `install.sh` / `install.ps1` are kept, as discussed in the issue — they only fire when the release API can't be resolved and are still correct there.

## How it was verified

Repo visibility confirmed public via `gh repo view` (`visibility: PUBLIC`). The `install.sh` change removes two comment lines only — no executable logic changed and the SHA256SUMS verification is untouched; `sh -n install.sh` passes. README change is a deleted blockquote, no commands affected.

## Checklist

- [ ] `cargo test --workspace` passes — N/A: this repo contains no Rust source; docs/installer comment only

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,13 +282,6 @@ you own or are authorised to administer.
 
 ## Installers
 
-> **While this repository is private**, the one-line installers and the release
-> download URLs return 404 for anonymous callers — GitHub does not serve private
-> content that way. Until it is made public, export a `GITHUB_TOKEN` with access
-> to this repository before running an installer, or fetch the archive with
-> `gh release download`.
-
-
 `install.sh` and `install.ps1` fetch and run executables, so here is exactly
 what they do:
 
--- a/install.sh
+++ b/install.sh
@@ -8,8 +8,6 @@
 # `npx skills add reolink/reolink-cli` instead (it fetches the binary for you).
 #
 # Env overrides: REOLINK_REPO=owner/repo, REOLINK_PREFIX=/install/root.
-# (While the repo is private, GitHub won't serve release assets over the public
-#  download URL — test with `gh release download` until it is public.)
 set -eu
 
 REPO="${REOLINK_REPO:-reolink/reolink-cli}"


reolink-net (approved):
Verified against the live repo before approving.

#6's claim checks out: the repository is public, yet the README and install.sh still told anonymous callers they would get a 404 and needed a GITHUB_TOKEN. The conditional fallback strings are correctly left in place.

Thanks for catching this — the drift was mine.

Issue #7, opened 2026-07-28
Automate the nine-file version sync: bump script + CI check against the latest release


ch-bas:
### What

The release version appears in nine places in this repo that must all agree, and must match the latest published release:

- `package.json`
- `.claude-plugin/plugin.json`
- `.claude-plugin/marketplace.json`
- `.codex-plugin/plugin.json`
- `.cursor-plugin/plugin.json`
- `gemini-extension.json`
- `openclaw.plugin.json`
- the README version badge
- `skills/reolink-cli/references/setup.md` (the `--version` example output)

CONTRIBUTING already states the rule ("Keep manifests in step … they must agree"), but nothing enforces it, and the failure it warns about just happened: **v0.10.1 shipped on 2026-07-27 and every one of those nine locations still said `0.10.0` two days later**, with no `[0.10.1]` CHANGELOG entry either (#3, PR #5).

### Proposal

Two small pieces — the check matters more than the script:

1. **`scripts/set-version.sh <version>`** — rewrites all nine locations in one go. Keep the file+pattern pairs in a single list inside the script so adding a tenth location is a one-line change. Refuse versions that don't look like `X.Y.Z`.

2. **A CI workflow (`.github/workflows/version-sync.yml`)** that fails when:
   - the nine locations disagree with *each other* (runs on every PR/push — catches a hand-edit that misses a file), or
   - they disagree with the **latest published release tag**, or the top `CHANGELOG.md` heading isn't that version (runs on `release: published` and on a daily schedule — catches the real failure mode above, where a release ships and nobody touches this repo at all).

The CHANGELOG *entry* stays human-written — it needs actual release-note content — so the check only asserts the top heading's version, and the script doesn't touch `CHANGELOG.md`.

### Caveats

- Releases are cut upstream from the private Rust source repo, so the *cleanest* long-term fix is upstream's release pipeline bumping this repo automatically (the 0.10.1 release notes mention that pipeline already gained a filename guard — this would be a natural sibling to it). That's an upstream change; the script + check here is the part that can land as a PR in this repo, and stays useful as a safety net even if upstream automates later.
- The scheduled check needs nothing but anonymous GitHub API access (the repo is public), so no secrets are involved.

I can open a PR with the script and workflow if this sounds right.

reolink-net:
The script half is in (#27, merged) — as `--set` on the existing `check-version-sync.sh` rather than a separate `set-version.sh`:

```bash
[...]

PR #8, opened 2026-07-28
Add a version-sync check; stop calling 0.6.0 "unreleased"

### LINKED ISSUE #3: CHANGELOG: `[0.6.0]` is still marked "unreleased" below seven shipped versions

reolink-net:
Closes #3. Part of #7.

## Why

Keeping the version in step across nine files has failed **twice**:

- 0.10.0 — the plugin manifests were left on the previous version
- 0.10.1 — this entire repository stayed a version behind while the release was already published, so the README badge showed `0.10.0` over a `0.10.1` download. @ch-bas caught that one in #5, not us.

## What

`scripts/check-version-sync.sh` resolves the latest published release and compares all nine locations plus the CHANGELOG heading, listing every file that disagrees.

Verified in both directions — it passes on the current tree, and it catches both real mistakes when reproduced:

```
MISMATCH  package.json                    0.10.0 (want 0.10.1)
MISMATCH  CHANGELOG.md                    no "## [0.10.1]" heading
```

The changelog check first used a prefix match, which happily accepted `## [0.10.1-rc]` as the entry for `0.10.1`. The negative-control test is what exposed it; it now anchors the closing bracket.

## On the bump script in #7

Not added. Rewriting nine strings is one `sed` and happens once per release; the failure mode is *forgetting*, not mistyping. A checker catches forgetting — a second script you must remember to run does not. Leaving #7 open for the CI half if you want it wired to Actions.

## Also

`[0.6.0]` was marked "unreleased" while sitting below seven shipped versions, which reads as a pending release (#3). It was prepared but never published here — 0.7.0 was the first — so it now says that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -249,9 +249,10 @@ Patch release. Fixes found in a pre-release audit, verified on real hardware.
   address works out of the box.
 - **Prebuilt binaries** — macOS arm64, Linux x86_64, Windows x86_64.
 
-## [0.6.0] — unreleased
+## [0.6.0] — not published
 
-Initial public distribution.
+Prepared as the first public distribution but never released here; the
+first published release was 0.7.0. Kept for the record of what it covered.
 
 - LAN-only camera operation: discover, login, info, snapshot, live preview, PTZ,
   IR / spotlight / LEDs, image tuning, OSD, motion + AI detection, recording +
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,8 +27,18 @@ you want. We implement it upstream and it arrives in the next release.
   the docs that had never been executed — a flag that did not exist, a
   subcommand that had been renamed. If you touch an example, run it and paste
   the output in the PR.
-- **Keep manifests in step.** The version appears in several manifests; they
-  must agree. Do not hand-edit a version to something no release uses.
+- **Keep manifests in step.** The version appears in nine places — seven
+  manifests, the README badge, and `skills/reolink-cli/references/setup.md` —
+  plus a CHANGELOG heading. They must all agree with the published release. Do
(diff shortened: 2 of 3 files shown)


reolink-net:
Reopening from a different account so it can be reviewed — the author of a PR cannot approve it.

Issue #15, opened 2026-07-29
install.ps1 does not fail loudly when the archive is missing the binaries (install.sh does)


ch-bas:
### What

When the extracted archive doesn't contain the expected binaries, the two installers behave differently:

- `install.sh` line 108 fails loudly: `[ -n "$src" ] || die "$b not found in the archive"`.
- `install.ps1` lines 69–70 pipe `Get-ChildItem … | Select-Object -First 1 | ForEach-Object { Copy-Item … }` — if nothing is found, the pipeline is empty, nothing is copied, and the script continues. It only dies later at line 83 when `& reolink-cli.exe config init` can't find the file, with an unrelated-looking error instead of "reolink-cli.exe not found in the archive".

Not exploitable (checksum verification has already passed by then), but for a security-sensitive script the failure should be explicit and name the actual problem, with parity between the two installers.

### Suggested fix

Resolve each binary to a variable first and `throw "<name> not found in the archive"` when the lookup comes back empty, mirroring install.sh.

PR #16, opened 2026-07-29
fix: purge on Windows used an env var uninstall.ps1 never reads

### LINKED ISSUE #12: `/reolink-cli:uninstall` resurrects the REOLINK_PURGE env var that 0.9.1 removed
### What

`commands/uninstall.md` (the `/reolink-cli:uninstall` slash command) tells Windows users to purge with an environment variable:

```powershell
$env:REOLINK_PURGE = "1"; .\uninstall.ps1
```

That variable does not exist. `uninstall.ps1` only forwards arguments and never reads the environment — this exact misdocumentation was already found and fixed once, in the 0.9.1 release:

> **`AGENTS.md` documented a `REOLINK_PURGE` environment variable for a full wipe on Windows.** `uninstall.ps1` only forwards arguments and never reads the environment, so anyone following it believed their stored credentials had been erased while the files remained on disk. Use `--purge` on both platforms. (CHANGELOG.md, 0.9.1)

`AGENTS.md` line 152 now explicitly says "There is no `REOLINK_PURGE` environment variable; the wrappers only forward arguments" — but the slash command file was never updated, so the bug survives on the surface where it does the most harm: a Claude Code user running `/reolink-cli:uninstall purge` on Windows gets a plain (non-purge) uninstall while being told their data was wiped. Camera credentials (`aliases.toml` + `credentials.key`) remain on disk.

### Also in the same file

The manual fallback (line 41) is `pkill -f reolink-gateway`, which kills **every** gateway process on the machine, including ones belonging to other installs — the exact over-reach that 0.9.1 fixed in `setup --uninstall` ("Termination is now scoped to the prefix being removed"). It also pattern-matches any command line containing that string. The fallback should target the prefix being removed, e.g. `pkill -f "$HOME/.local/bin/reolink-gateway"`.

### Suggested fix

- Replace the Windows purge variant with `.\uninstall.ps1 --purge` (matching AGENTS.md and the 0.9.1 changelog).
- Scope the `pkill` fallback to the install prefix.

Verified against the real v0.10.1 release scripts: `uninstall.ps1` in the tarball only forwards `$args`.

ch-bas:
## What changes

Fixes #12. The `/reolink-cli:uninstall` slash command told Windows users to purge with `$env:REOLINK_PURGE = "1"` — an environment variable `uninstall.ps1` never reads (it only forwards arguments), so the user got a plain uninstall while being told their credentials were wiped. This exact misdocumentation was fixed in AGENTS.md for 0.9.1; the slash command file was missed. Now `.\uninstall.ps1 --purge` on both platforms.

Also scopes the manual fallback from `pkill -f reolink-gateway` (kills gateways from every install on the machine) to the full binary path of the prefix being removed, matching the prefix-scoping 0.9.1 established for `setup --uninstall`.

## How it was verified

Downloaded the real v0.10.1 release tarball (checksum-verified against its `SHA256SUMS`) and read its bundled `uninstall.ps1`: it only forwards `$args`, confirming the env var is never read. Cross-checked against AGENTS.md line 152 and the 0.9.1 CHANGELOG entry, which both state `--purge` is the only mechanism.

## Checklist

### DIFF
--- a/commands/uninstall.md
+++ b/commands/uninstall.md
@@ -31,14 +31,15 @@ The release tarball ships its own `uninstall.sh` / `uninstall.ps1`. Run them fro
 **Windows — purge:**
 
 ```powershell
-$env:REOLINK_PURGE = "1"; .\uninstall.ps1
+.\uninstall.ps1 --purge
 ```
 
 If the user no longer has the tarball on disk, falling back to manual removal works too:
 
 ```bash
-# macOS / Linux
-pkill -f reolink-gateway
+# macOS / Linux — kill only the gateway from this install prefix,
+# never gateways belonging to another installation
+pkill -f "$HOME/.local/bin/reolink-gateway"
 rm -f ~/.local/bin/reolink-cli ~/.local/bin/reolink-gateway
 # (purge only)
 rm -rf ~/.config/reolink-cli ~/.cache/reolink-cli ~/.local/state/reolink-cli


reolink-net:

PR #17, opened 2026-07-29
fix: cache-clean slash command named a cache directory that does not exist

### LINKED ISSUE #13: `/reolink-cli:cache-clean` names the wrong cache directory (`~/.cache/reolink/`)
### What

`commands/cache-clean.md` (the `/reolink-cli:cache-clean` slash command) names the wrong cache directory in both its frontmatter description and its body:

- line 2: "Preview + clean artifacts under `~/.cache/reolink` older than 7 days"
- line 6: "Clean old cache entries under `~/.cache/reolink/` …"

The real directory is `~/.cache/reolink-cli/` — every other reference agrees (SKILL.md, README.md, AGENTS.md), and the binary confirms it: `reolink-cli cache status` on v0.10.1 reports `root: ~/Library/Caches/reolink-cli` (macOS) / `~/.cache/reolink-cli` (Linux).

Harmless to the command itself (the CLI resolves its own path), but an agent or user reading the description and then inspecting/deleting the named directory by hand targets a path that doesn't exist — or worse, a different tool's directory if one ever occupies `~/.cache/reolink/`.

### Suggested fix

`~/.cache/reolink/` → `~/.cache/reolink-cli/` in both places.

ch-bas:
## What changes

Fixes #13. The `/reolink-cli:cache-clean` slash command named `~/.cache/reolink/` in its frontmatter description and body; the CLI's cache root is `~/.cache/reolink-cli/`. The command itself was unaffected (the CLI resolves its own path), but anyone inspecting or deleting the named directory by hand was pointed at the wrong place.

## How it was verified

Ran `cache status` on the actual v0.10.1 binary (downloaded from the release, checksum-verified): it reports `root: ~/Library/Caches/reolink-cli` on macOS (`~/.cache/reolink-cli` on Linux). All other docs (SKILL.md, README.md, AGENTS.md) already use the `-cli` path.

## Checklist

### DIFF
--- a/commands/cache-clean.md
+++ b/commands/cache-clean.md
@@ -1,9 +1,9 @@
 ---
-description: Preview + clean artifacts under ~/.cache/reolink older than 7 days
+description: Preview + clean artifacts under ~/.cache/reolink-cli older than 7 days
 argument-hint: [--older-than 30d]
 ---
 
-Clean old cache entries under `~/.cache/reolink/` (snapshots, TTS PCM clips, etc.).
+Clean old cache entries under `~/.cache/reolink-cli/` (snapshots, TTS PCM clips, etc.).
 
 **Step 1 — preview (dry-run, always first):**
 


reolink-net:

PR #18, opened 2026-07-29
fix: AGENTS.md pointed agents at a version: frontmatter SKILL.md does not have

### LINKED ISSUE #14: AGENTS.md Part 5 points agents at a `version:` frontmatter that SKILL.md does not have
### What

`AGENTS.md` Part 5 (line 192) instructs agents:

> If the binary version on disk and the version named in `SKILL.md` (`version:` frontmatter, see the local skill file) disagree, tell the user …

`SKILL.md` has no `version:` frontmatter key — its frontmatter is `name`, `description`, and `metadata` only. An agent following Part 5 goes looking for a field that can never be found.

### Suggested fix

The binary already provides the intended check, machine-readably: `reolink-cli features --output json` returns `.data.version`, `.data.claudeCodePluginVersion`, and `.data.cacheState` (values like `no_plugin_cache` / `in_sync`), and SKILL.md itself already documents "run `plugin refresh` when `features` reports `cache_state != in_sync`". Point Part 5 at those fields instead of the nonexistent frontmatter key.

Verified against the v0.10.1 binary: `features --output json` contains all three fields; SKILL.md frontmatter contains no `version:`.

ch-bas:
## What changes

Fixes #14. AGENTS.md Part 5 told agents to compare the binary version against "the version named in `SKILL.md` (`version:` frontmatter)" — a field SKILL.md has never had (its frontmatter is `name`, `description`, `metadata`). The paragraph now points at the check the binary actually provides: `features --output json` → `.data.version` (binary), `.data.claudeCodePluginVersion` (loaded skill tree, `null` when not a Claude Code plugin), and `.data.cacheState`, with `plugin refresh` as the remedy — matching what SKILL.md already prescribes for `cache_state != in_sync`.

## How it was verified

Ran `features --output json` on the actual v0.10.1 binary (downloaded from the release, checksum-verified): all three fields are present (`version: "0.10.1"`, `claudeCodePluginVersion: null`, `cacheState: "no_plugin_cache"` on a machine without the plugin installed). Confirmed SKILL.md's frontmatter keys by reading the file.

## Checklist

- [ ] `cargo test --workspace` passes — N/A: this repo contains no Rust source; docs only

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -189,7 +189,7 @@ reolink-cli cache status --output json 2>/dev/null \
 
 **Read `.data.commands`** from the `features` output before dispatching any subcommand. If a command in SKILL.md is missing from that list, the installed binary is older than the skill and the right response is to tell the user to upgrade by re-extracting their newest release tarball over the old one.
 
-If the binary version on disk and the version named in `SKILL.md` (`version:` frontmatter, see the local skill file) disagree, tell the user — the skill might be describing capabilities the binary doesn't have yet (or vice versa). Both should match the version printed by `reolink-cli --version`.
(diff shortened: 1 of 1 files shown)


reolink-net:

Issue #10, opened 2026-07-28
Channels ≥ 11 unreachable on RLN8-410: "route not found: unable to detect protocol" while ch10 works in the same session


Pharkie:
## Summary

On an RLN8-410 NVR whose channels are **non-contiguous** (`1, 9, 10, 11`), every command targeting **channel 11** fails with a device-level routing error, while channel 10 succeeds in the same session against the same host:

```
route not found: unable to detect protocol for '<NVR-IP>' — device unreachable, or not a Reolink v20/v30 device
```

The host is demonstrably reachable and the camera on channel 11 is healthy — Home Assistant's `reolink` integration (`reolink-aio` 0.21.4) drives it fine at the same moment.

## Environment

| | |
|---|---|
| reolink-cli | `0.10.1 (external · LAN-only)` |
| Install | `install.sh` → `~/.local/bin` |
| Host OS | macOS 15 (arm64) |
| NVR | RLN8-410, hardware `N7MB01`, firmware `v3.6.5.562_26062933` |
| Gateway | `reolink-gateway 127.0.0.1:9000` (running) |
| Transport | v20 / TCP 9000, direct `--host` (different VLAN from the client, unicast routed) |

Channel list `[1, 9, 10, 11]` confirmed independently via the Home Assistant reolink integration's diagnostics (`data.channels`).

## Reproduction

Registry entries are identical apart from `channel`:

```toml
[cameras.ch10cam]
host = "<NVR-IP>:9000"
user = "<user>"
channel = 10

[cameras.ch11cam]
host = "<NVR-IP>:9000"
user = "<user>"
channel = 11
```

A/B/A within a single session, immediately after `reolink-cli --camera nvr login`:

```console
$ reolink-cli --camera ch10cam osd get
{"ok":true,"command":"osd","protocol":"v20","data":{"datetimeEnable":true,"name":"<redacted>","nameEnable":true},"error":null}

$ reolink-cli --camera ch11cam osd get
{"ok":false,"command":"osd","protocol":null,"data":null,"error":{"code":"NETWORK_ERROR","message":"route not found: unable to detect protocol for '<NVR-IP>' — device unreachable, or not a Reolink v20/v30 device","retryable":true}}

$ reolink-cli --camera ch10cam osd get
{"ok":true,"command":"osd","protocol":"v20","data":{"datetimeEnable":true,"name":"<redacted>","nameEnable":true},"error":null}
```

[...]

Issue #9, opened 2026-07-28
[FR] Allow fetching of system stats (CPU, Memory etc)


second2falcon:
### What are you trying to do?

Please allow the fetching of system statistics like CPU usage, memory utilization etc. 

See reactions on Reddit post:

https://www.reddit.com/r/reolinkcam/s/xFkOVmGJjm

As soon as this is available I am happy to help test it due to me having issues with stuttering ONVIF streams (according to Home Assistant) due to high CPU and memory usage. 

### How do you do it today?

Via Home Assistant (which also brings its number of issues)

### Camera models this applies to

All (but in my case Doorbell WiFi)

reolink-net:
Shipped in [v0.10.2](https://github.com/reolink/reolink-cli/releases/tag/v0.10.2).

```console
$ reolink-cli --camera front-door config get performance
{"ok":true,"command":"config","protocol":"v20",
 "data":{"channel":0,"path":"performance",
         "value":{"cpuUsedPercent":60,"codeRate":0,"netDataRate":8}}}
```

That output is from a real E1 Outdoor a few minutes ago, using the published 0.10.2 binary.

## What you get

| Field | Meaning |
|---|---|
| `cpuUsedPercent` | CPU load, 0–100 |
| `codeRate` | Encoder bitrate — reads 0 on an idle camera, meaningful while a stream runs |
| `netDataRate` | Network throughput |

**No memory figure.** The device does not report one — this is everything the firmware exposes, and I would rather say that than invent a field.

## Two caveats worth knowing before you test

**Not every model implements it.** Unsupported firmware answers with a device-level rejection rather than a zero. A camera reported as 0% busy when it never said so would be worse than an error, so there is no fallback value. Of the units I could reach: an E1 Outdoor answers, an OMVI X16 refuses.

**It is instantaneous, not averaged.** Two calls a second apart legitimately differ. For your stuttering-stream investigation you probably want to sample it:

```bash
while :; do
  printf '%s ' "$(date +%T)"
  reolink-cli --camera doorbell config get performance --output json | jq -c .data.value
  sleep 5
done
```

Correlating `cpuUsedPercent` against the moments Home Assistant reports stutter is the comparison that would actually implicate CPU.

## If you test it

You mentioned a Doorbell WiFi. I could not verify that model — mine was asleep every time I tried. If it answers, the numbers during a stutter would be genuinely useful; if it refuses, tell me and I will note the model as unsupported rather than leaving you guessing.

[...]

PR #21, opened 2026-07-29
docs: sync the distribution to the 0.10.2 release

cuican-reolink:
Adds the `[0.10.2]` CHANGELOG entry and moves all nine version locations.

Verified with the checker added in #11 — the one added because this same sync was missed by hand twice:

```
$ ./scripts/check-version-sync.sh
latest published release: 0.10.2
  ok        package.json                                   0.10.2
  ...
  ok        CHANGELOG.md                                   has an entry

All version strings agree with 0.10.2.
```

The release itself (#9 `config get performance`, #10 probe error message) is already published.


reolink-net (approved):
Verified: `./scripts/check-version-sync.sh` passes against the published 0.10.2, and the CHANGELOG entry matches the release notes.

Issue #20, opened 2026-07-29
Security reporting is not straightforward: the official support form has no path for software vulnerabilities


ch-bas:
> **Note:** this issue is about the *reporting channel only*. It does not describe any vulnerability, and per SECURITY.md none should be posted in this thread.

### What

SECURITY.md routes vulnerability reports to [Reolink support](https://support.reolink.com), with the instruction to "state that the report concerns a **security issue in `reolink-cli`**, so it is routed rather than handled as a normal support ticket". The issue-template config (`.github/ISSUE_TEMPLATE/config.yml`) points its "Security vulnerability" contact link at the same portal.

But that portal has no place for this kind of report. Its visible categories are consumer product support — setup, troubleshooting, password/account help, recording & alerts, firmware, orders & payment, warranty & returns — and the only path in is a generic "Contact Support" request form. There is no category, form field, or documented route for a security vulnerability or a technical bug in a software tool.

In practice that means a researcher's vulnerability details enter a consumer ticket queue, triaged by support agents who handle camera and warranty questions and may never have heard of `reolink-cli`. The friction cuts both ways: it discourages the reports SECURITY.md asks for, and it increases the odds a report is mishandled, misrouted — or just posted as a public issue instead.

### Suggested fix

**Enable GitHub Private Vulnerability Reporting on this repository.** It is currently disabled (`GET /repos/reolink/reolink-cli/private-vulnerability-reporting` → `{"enabled": false}`), and it's a one-toggle setting (Settings → Advanced Security → Private vulnerability reporting). Once on, researchers get a private "Report a vulnerability" button on the repo's Security tab — visible only to maintainers, no consumer support queue in the loop, and it's the mechanism GitHub-native reporters look for first.

Then:

- **SECURITY.md** — point "Reporting a vulnerability" at the Security tab / PVR flow, keeping the support portal only for the already-out-of-scope firmware and mobile/desktop-app issues.
- **`.github/ISSUE_TEMPLATE/config.yml`** — point the "Security vulnerability" contact link at `https://github.com/reolink/reolink-cli/security/advisories/new` instead of the support portal.

Enabling PVR is a repo-settings action only maintainers can take, which is why this is an issue rather than a PR. Happy to follow up with the small SECURITY.md + `config.yml` PR once it's enabled.

reolink-net:
Closing as done, in the order that made it real:

[...]

PR #23, opened 2026-07-29
Route security reports to GitHub's private advisory flow

### LINKED ISSUE #15: install.ps1 does not fail loudly when the archive is missing the binaries (install.sh does)

cuican-reolink:
Closes #15, #20.

## #15 — reproduced, fixed, verified with real PowerShell

The old pipeline, run under PowerShell 7.6.4 with nothing to find:

```
>>> pipeline finished, script carried on
>>> bin contains: 0 files
exit code: 0
```

After the fix, same input: `reolink-cli.exe not found in the archive`. Normal path still copies both binaries, and the file re-parses clean via `[Parser]::ParseFile`.

**Scope note:** the installer *inside* the release archive is a different script (359 lines vs 101) and was already correct — it uses `Test-Path` and `exit 1`. Only the one-line curl installer in this repository had the gap. I checked rather than assuming they shared code.

## #20 — routed to the private advisory flow

Both `SECURITY.md` and the issue-template contact link sent vulnerability reports to the consumer support portal, which has no category for a software vulnerability. Both now point at `security/advisories/new` (verified reachable). The support portal is kept as a fallback with its limitation **stated rather than hidden**, since a reader who cannot use GitHub still needs somewhere to go.

⚠️ **This PR does not enable Private Vulnerability Reporting.** `GET /repos/.../private-vulnerability-reporting` still returns `{"enabled": false}` — that is an admin-only setting. Until a maintainer flips it (Settings → Advanced Security), the link shows researchers a page they cannot submit from. **Merging without that toggle points the docs somewhere non-functional.**

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,8 +1,8 @@
 blank_issues_enabled: false
 contact_links:
   - name: Security vulnerability
-    url: https://support.reolink.com
-    about: Do not open a public issue. Report privately — see SECURITY.md.
+    url: https://github.com/reolink/reolink-cli/security/advisories/new
+    about: Do not open a public issue. Report privately — visible only to maintainers.
   - name: Camera or account help (not this tool)
     url: https://support.reolink.com
     about: Firmware, warranty, app or account questions belong with Reolink support.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -5,9 +5,16 @@
 Please report security issues **privately** — do not open a public issue, and
 do not describe the problem in a pull request.
 
-Contact [Reolink support](https://support.reolink.com) and state that the report
-concerns a **security issue in `reolink-cli`**, so it is routed rather than
-handled as a normal support ticket.
+Use GitHub's private reporting:
+**[Report a vulnerability](https://github.com/reolink/reolink-cli/security/advisories/new)**
+(also reachable from the repository's Security tab). The report is visible only
+to maintainers, and the discussion stays on the advisory until a fix ships.
+
+If that page is unavailable to you, contact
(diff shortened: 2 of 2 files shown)


reolink-net:
Holding this until Private Vulnerability Reporting is enabled.

Checked again just now: `GET /repos/reolink/reolink-cli/private-vulnerability-reporting` → `{"enabled": false}`, and `PUT` returns 404 for a maintainer — it needs admin.

Merging the #20 half before that toggle would point `SECURITY.md` and the issue-template link at a page a researcher cannot submit from. That is worse than the current state, where at least the support portal accepts *something*.

**@reolink-net / maintainers:** Settings → Advanced Security → Private vulnerability reporting → Enable. One toggle; then this merges as-is.

The #15 half (install.ps1) has no such dependency and is verified — happy to split it out if you would rather ship that now and leave the docs pending.

ch-bas:
One merge-order caveat: as of today, Private Vulnerability Reporting is still **disabled** on this repo — `GET /repos/reolink/reolink-cli/private-vulnerability-reporting` returns `{"enabled": false}`. While it's off, `https://github.com/reolink/reolink-cli/security/advisories/new` works only for users with maintainer access; everyone else gets a 404.

So if this merges before the setting is flipped, SECURITY.md and the issue-template link will point reporters at a dead page — arguably worse than the support-portal status quo, since there's no error message explaining why.

Suggested order: enable the toggle first (Settings → Advanced Security → Private vulnerability reporting), verify the link loads from a logged-out/non-maintainer session, then merge this. The doc changes themselves look right to me, including keeping the support portal as the explicit fallback. (Context: #20.)

reolink-net (approved):
Private vulnerability reporting is enabled now, so this points researchers at a page that accepts submissions. Held until that was true — merging first would have documented a door that did not open.

PR #24, opened 2026-07-29
install.ps1: fail by name when the archive is missing a binary

### LINKED ISSUE #15: install.ps1 does not fail loudly when the archive is missing the binaries (install.sh does)

cuican-reolink:
Closes #15. Split out of #23 so it does not wait on the admin-only toggle that half needs.

Reproduced under PowerShell 7.6.4 — old pipeline with nothing to find exits **0** with an empty `bin/`. After the fix, same input throws `reolink-cli.exe not found in the archive`. Normal path still copies both; the file re-parses clean via `[Parser]::ParseFile`.

**Scope note:** the installer inside the release archive is a different script (359 lines vs 101) and was already correct — `Test-Path` + `exit 1`. I checked rather than assuming they shared code; only this one had the gap.


reolink-net (approved):
Verified with PowerShell 7.6.4 before approving: the old pipeline exits 0 on an empty result, the new loop throws by name, and the normal path is unaffected.

Issue #25, opened 2026-07-29
device expand fails on RLN8-410: GetSupport returns code 300


Pharkie:
`device expand` fails on an RLN8-410, which is inside its documented scope. The help says "Scope: RLN-series v20 NVR only", and login on this device reports protocol `v20`.

```
$ reolink-cli device expand nvr --gateway-addr 127.0.0.1:9000
{"ok":false,"error":{"code":"PROTOCOL_ERROR",
 "message":"backend error: device error: device command 199 failed (code=300)"}}
```

Three consecutive runs, same error each time. `capabilities` fails identically, so this looks like the firmware declining `GetSupport` rather than anything specific to expand.

Environment: RLN8-410, firmware `v3.6.5.562_26062933`, hardware `N7MB01`, reolink-cli 0.10.1, LAN only.

## Why a fallback is awkward here

Channels have to be registered by hand, and guessing `0..N` doesn't help: the four cameras on this NVR are on channels **1, 9, 10 and 11**.

## One thing that does work

`GetRtspUrl` answers per channel and returns real paths:

| Channel | RTSP path |
|---|---|
| 1 | `Preview_02_main` |
| 9 | `Preview_10_main` |
| 10 | `Preview_11_main` |
| 11 | `Preview_12_main` |

They're zero padded and don't match the channel numbers. That's how I mapped these four, so a `GetRtspUrl`-based fallback looks possible when `GetSupport` is unavailable.

Happy to test a patch on this hardware.

reolink-net:
Fixed — and your report turned up a second bug I would not have found otherwise.

## Two problems, not one

**1. cmd 199 refused.** Your firmware answers code 300, and that aborted the whole command.

**2. Even when 199 works, `expand` was writing the wrong channels.** `channelNum` is a *count*, and the code registered `0..count`. Your cameras are on **1, 9, 10, 11** — so on a working NVR it would have created four entries pointing at slots 0–3, three of them empty. Your note that the channels are non-contiguous is what exposed this; a report that only said "199 fails" would have left it in place.

## What it does now

199 is asked for an upper bound when it answers, and ignored when it does not. Then each channel is probed, and **only the ones that answer are registered, with their real channel number**. 199 no longer decides *which* channels exist — only how far to look.

The probe uses **cmd 44 (OSD get)**, which is per-channel and refused on an empty slot — 400 on your NVR, 421 on a single-camera unit. That is your own observation from #10, and it is already implemented, so this needs no new protocol work.

[...]

PR #26, opened 2026-07-29
release: 0.10.3

cuican-reolink:
Publishes the 0.10.3 release: `device expand` now probes each channel instead of trusting `GetSupport` (#25), and the nine version strings move to 0.10.3.

`./scripts/check-version-sync.sh 0.10.3` passes on all nine locations.


reolink-net (approved):
Version sync verified across all nine locations.

PR #27, opened 2026-07-29
check-version-sync: add --set to rewrite all nine locations

cuican-reolink:
Closes half of #7 — the `set-version.sh` half. The CI half is a decision for the repo owners; see below.

## What changed

`check-version-sync.sh` gained `--set <version>`, which rewrites all nine locations:

```console
$ ./scripts/check-version-sync.sh --set 0.11.0
setting every version string to 0.11.0

  ok        package.json                                   0.11.0
  ok        openclaw.plugin.json                           0.11.0
  ok        gemini-extension.json                          0.11.0
  ok        .claude-plugin/plugin.json                     0.11.0
  ok        .claude-plugin/marketplace.json                0.11.0
  ok        .codex-plugin/plugin.json                      0.11.0
  ok        .cursor-plugin/plugin.json                     0.11.0
  ok        README.md                                      0.11.0
  ok        skills/reolink-cli/references/setup.md         0.11.0
  MISMATCH  CHANGELOG.md                                   no "## [0.11.0]" heading (write it by hand)
```

## Why it went in the existing script rather than a new `set-version.sh`

The issue asks for the file list to live in "a single list inside the script". Two scripts means two lists that must agree — the same failure class the whole issue is about. So there is one `LOCATIONS` list and both the check and the rewrite read it. Adding a tenth location is still one line.

## `--set` checks its own work

After writing, it falls through to the check rather than reporting success on its own. That matters for the one way a rewrite can fail quietly: a file whose pattern drifted. Verified by deleting the README badge line and running `--set`:

```
MISMATCH  README.md    <not found> (want 9.9.9)
```

Silently skipping there would be worse than not having the script — you would believe nine files were bumped when eight were.

## `--set` will not resolve the version from the release

It requires an explicit `X.Y.Z`. You run it while preparing a version that is *not* published yet; stamping "whatever is latest" is how you ship the previous number.

Version validation is exactly three numeric segments. Both of these were accepted by looser drafts before I tightened it, and either would have been stamped into all nine files without complaint:

| input | result |
|---|---|
| `1.2` | rejected (parsed as major=1 minor=2 patch=2 by the first draft) |
| `1.2.3.4` | rejected |
| `v1.2.3` | rejected |
| `0.11.0` | accepted |

## Tested

On a scratch copy of this repo: all nine rewritten and independently re-grepped; all seven JSON manifests still parse under `python3 -m json.tool`; running `--set` twice leaves the files byte-identical; the check against the real repo still passes for the published 0.10.3.

## The CI half is deliberately not here

A workflow is the only thing that catches the failure mode the issue actually names — *"a release ships and nobody touches this repo at all."* A local script does not, because it only helps when someone remembers to run it.

I am not adding it in this PR because it would be this repository's first GitHub Actions workflow, which is a supply-chain decision for the repo owners rather than something to slip in alongside a shell script. Leaving #7 open for that.

Two things I looked at and rejected rather than leave unmentioned:

- **A version gate in the upstream packaging script.** Builds happen *before* this repo is synced, so it would fire on every legitimate release. A gate that cries wolf gets ignored, which is worse than no gate.
- **A `release.sh` owning the whole publish flow.** It would have to wrap release creation, deletion of superseded releases, and cross-account PRs — irreversible, outward-facing steps that should stay deliberate.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,15 +30,22 @@ you want. We implement it upstream and it arrives in the next release.
 - **Keep manifests in step.** The version appears in nine places — seven
   manifests, the README badge, and `skills/reolink-cli/references/setup.md` —
   plus a CHANGELOG heading. They must all agree with the published release. Do
-  not hand-edit a version to something no release uses. Check with:
+  not hand-edit a version to something no release uses.
 
   ```bash
-  ./scripts/check-version-sync.sh
+  ./scripts/check-version-sync.sh              # do they agree with the latest release?
+  ./scripts/check-version-sync.sh --set 0.11.0 # rewrite all nine, then check
   ```
 
-  Doing this by hand has failed twice: once the plugin manifests were left
+  Doing this by hand has failed three times: once the plugin manifests were left
   behind, once the whole repository stayed a version back while the release was
-  already published, so readers saw a stale badge over a newer download.
+  already published — readers saw a stale badge over a newer download — and once
+  more while cutting 0.10.3. Use `--set`; it writes every location from one list
+  and then checks its own work, so a location it could not match is reported
+  rather than skipped.
+
+  `--set` does not touch `CHANGELOG.md`. That entry needs release notes a script
+  cannot write, so the check reports it as missing until you write it.
 - **Never commit real data.** No camera passwords, tokens, UIDs, serial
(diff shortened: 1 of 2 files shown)


reolink-net (approved):
Tested on a scratch copy; the missing-pattern case reports rather than skips, which is the property that matters.

PR #28, opened 2026-07-29
release: 0.10.4

cuican-reolink:
Publishes 0.10.4: four fixes found by installing 0.10.3 and running it against real hardware, the first being 0.10.3's own `device expand` naming bug.

Version strings bumped with the `--set` mode from #27 — its first real use. One command instead of nine hand-edits, and the CHANGELOG was correctly reported as missing until written by hand.

`./scripts/check-version-sync.sh` passes against the published release.


reolink-net (approved):
Version sync verified against the published 0.10.4.

PR #31, opened 2026-07-29
fix: scope the uninstall fallback pkill to the install prefix

### LINKED ISSUE #29: uninstall.md manual fallback still kills gateways from every install (`pkill -f` leftover from #12)
### What

#22 fixed the `REOLINK_PURGE` half of #12, but the second finding in that issue survived: the manual fallback in `commands/uninstall.md` (line 41) is still

```bash
pkill -f reolink-gateway
```

`pkill -f` matches the **full command line of every process on the machine**, so this kills gateways belonging to other installations — the exact over-reach the 0.9.1 release fixed inside `setup --uninstall` ("Termination is now scoped to the prefix being removed"). It also pattern-matches any unrelated process whose command line merely contains the string.

Since the fallback in this file removes `~/.local/bin/` binaries specifically, the kill should be scoped the same way:

```bash
pkill -f "$HOME/.local/bin/reolink-gateway"
```

### Why it matters

This is the fallback an AI agent runs verbatim when the tarball is gone. An agent following the current line on a machine with a second install (e.g. a dev copy elsewhere on disk) takes down the wrong gateway silently.

PR follows.

ch-bas:
## What changes

Fixes #29. The manual fallback in `commands/uninstall.md` ran `pkill -f reolink-gateway`, which matches the full command line of every process on the machine — killing gateways from other installations (the over-reach 0.9.1 fixed inside `setup --uninstall`) and any unrelated process whose command line contains the string. #22 fixed the `REOLINK_PURGE` half of #12; this is the remaining half. The kill is now scoped to the same `~/.local/bin` prefix whose binaries the fallback removes.

## How it was verified

Behavioral claim checked against `pkill(1)` semantics: `-f` matches the full argument list, so the full-path pattern only matches a gateway started from `$HOME/.local/bin/`. Doc-only change; no camera interaction involved.

## Checklist

### DIFF
--- a/commands/uninstall.md
+++ b/commands/uninstall.md
@@ -37,8 +37,9 @@ The release tarball ships its own `uninstall.sh` / `uninstall.ps1`. Run them fro
 If the user no longer has the tarball on disk, falling back to manual removal works too:
 
 ```bash
-# macOS / Linux
-pkill -f reolink-gateway
+# macOS / Linux — kill only the gateway from this install prefix,
+# never gateways belonging to another installation
+pkill -f "$HOME/.local/bin/reolink-gateway"
 rm -f ~/.local/bin/reolink-cli ~/.local/bin/reolink-gateway
 # (purge only)
 rm -rf ~/.config/reolink-cli ~/.cache/reolink-cli ~/.local/state/reolink-cli


reolink-net (approved):
Merged. The over-reach is real and your ordering is the project's own: a stray process is recoverable, killing someone else's gateway is not.

Two things I checked before merging, because the new pattern has a gap and chasing it found a worse bug underneath.

**The full-path pattern misses a gateway started from `PATH`.** `pkill -f` matches the command line, and the command line is whatever was typed. Both of these are gateways I started on this machine:

```
94380 reolink-gateway 127.0.0.1:9596          <- started as `reolink-gateway ...` with ~/.local/bin on PATH
94381 ./target/release/reolink-gateway ...    <- a second install
```

`pgrep -f "$HOME/.local/bin/reolink-gateway"` matches neither. So the fallback is safe but often a no-op, and then `rm -f` deletes the binary out from under a live process. I merged anyway — failing to stop is the failure this project prefers — but the snippet is now a last resort that quietly does less than it looks like.

**The real defect is upstream of the doc.** The fallback only exists because `reolink-cli setup --uninstall` might not be available. That code path resolves each candidate's *executable* rather than matching its command line, which is the right approach and immune to how the process was launched — except that on macOS it reads the executable with `ps -o comm=`, under a comment asserting that prints the full path there. It does not. It prints argv[0]:

```
$ ps -o comm= -p 94380
reolink-gateway
```

`is_owned_by_prefix` rejects anything non-absolute, so on macOS `setup --uninstall` **silently skips the user's own gateway** whenever it was started by bare name, prints "runs from reolink-gateway — not this install", and then removes the binary. Same class as what you reported, one layer down, and it was in a comment nobody had run.

Fixing that in the CLI is where this belongs; it will ship in the next release. Thanks — the doc bug was worth reporting on its own, and it would not have surfaced the code bug if you had not been specific about *why* `-f` was the wrong matcher.

PR #32, opened 2026-07-29
fix: drop the dangling sentence left by the AGENTS.md Part 5 rewrite

### LINKED ISSUE #30: AGENTS.md Part 5 rewrite left a dangling sentence: "--version should match --version"
### What

The AGENTS.md paragraph rewritten in #22 (Part 5, line 192) now ends with a sentence left over from the text it replaced:

> The skill carries no version of its own — it is distributed with the release archive, so its version is that release. Compare `reolink-cli --version` against the latest release instead. **Both should match the version printed by `reolink-cli --version`.**

The bolded sentence made sense in the old text (where "both" were the binary version and a `version:` frontmatter), but after the rewrite "both" has no referent, and the sentence reduces to "`--version` should match `--version`". An agent reading Part 5 literally has nothing actionable to do with it.

### Suggested fix

Delete the dangling sentence — the preceding one already states the whole check ("Compare `reolink-cli --version` against the latest release").

PR follows.

ch-bas:
## What changes

Fixes #30. The AGENTS.md Part 5 rewrite in #22 kept the closing sentence of the text it replaced — "Both should match the version printed by `reolink-cli --version`." — whose "both" (binary version + the nonexistent `version:` frontmatter) no longer exists. After the rewrite it reduces to "`--version` should match `--version`". The preceding sentence already states the whole check, so the leftover is deleted.

## How it was verified

Read in context: the paragraph now ends at "Compare `reolink-cli --version` against the latest release instead.", which is complete and actionable on its own. Doc-only, one line removed.

## Checklist

- [ ] `cargo test --workspace` passes — N/A: this repo contains no Rust source; docs only

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -189,7 +189,7 @@ reolink-cli cache status --output json 2>/dev/null \
 
 **Read `.data.commands`** from the `features` output before dispatching any subcommand. If a command in SKILL.md is missing from that list, the installed binary is older than the skill and the right response is to tell the user to upgrade by re-extracting their newest release tarball over the old one.
 
-The skill carries no version of its own — it is distributed with the release archive, so its version is that release. Compare `reolink-cli --version` against the latest release instead. Both should match the version printed by `reolink-cli --version`.
+The skill carries no version of its own — it is distributed with the release archive, so its version is that release. Compare `reolink-cli --version` against the latest release instead.
 
 ---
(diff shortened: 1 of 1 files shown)


reolink-net (approved):
Correct — the second sentence restates the first with the same command on both sides, left over from the rewrite.

PR #33, opened 2026-07-30
release: 0.10.5, and a --self mode for the version check

cuican-reolink:
Two things: the 0.10.5 sync, and the script groundwork for the CI half of #7.

## 0.10.5

[Release notes](https://github.com/reolink/reolink-cli/releases/tag/v0.10.5). `setup --uninstall` did not stop your own gateway on macOS and then deleted the binary out from under it — the code-level version of the `pkill` over-reach reported in #29 and fixed in the docs by #31. Plus hardware verification of the #25 fix on an RLN4E running the reporter's exact firmware (v3.6.5.562).

Version strings bumped with `--set`, so the nine locations moved in one command and the CHANGELOG was correctly reported as missing until written by hand.

## `--self`

`check-version-sync.sh --self` asks only whether the nine locations agree **with each other**, taking `package.json` as the reference:

```
$ ./scripts/check-version-sync.sh --self
checking internal agreement against package.json: 0.10.5
  ok  package.json  0.10.5
  ...
```

It deliberately does not consult the published release. That is a different question, and a branch may legitimately carry a version no release uses yet — failing pull requests for that would train people to ignore the check.

Verified it reports rather than passes when a manifest drifts:

```
$ sed -i 's/"version": "0.10.5"/"version": "0.9.9"/' .cursor-plugin/plugin.json
$ ./scripts/check-version-sync.sh --self
  MISMATCH  .cursor-plugin/plugin.json    0.9.9 (want 0.10.5)
```

`--self` with a version argument, and `--self --set`, are both refused rather than silently doing something surprising.

## The workflow itself is not in this PR

It was, and I took it out. Pushing `.github/workflows/version-sync.yml` is rejected:

```
! [remote rejected] refusing to allow an OAuth App to create or update workflow
  `.github/workflows/version-sync.yml` without `workflow` scope
```

None of the tokens available to me carry that scope, and the contents API is under the same restriction, so this is not something I can work around — it needs a token authorised for workflows, or the file added through the web UI.

I split it rather than hold the release sync behind it: this repository currently shows a 0.10.4 badge over a 0.10.5 download, which is exactly the failure #7 is about, and there is no reason to leave it there while the CI question is sorted out.

The workflow is written and tested (two mutually exclusive jobs — `--self` on PR/push, the release comparison on `release: published` plus a daily schedule; `permissions: contents: read`; a settle delay so a just-published release does not report a mismatch that fixes itself). It goes in as a follow-up the moment the scope exists.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
     {
       "name": "reolink-cli",
       "description": "Operate Reolink cameras locally via reolink-cli: discover, login, inspect, configure, preview, snapshot, and control PTZ/light/audio/detection/recording.",
-      "version": "0.10.4",
+      "version": "0.10.5",
       "source": "./",
       "author": {
         "name": "Reolink"
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "reolink-cli",
   "description": "Operate Reolink cameras locally: discover, login, inspect, configure, preview, snapshot, and control PTZ/light/audio/detection/recording.",
-  "version": "0.10.4",
+  "version": "0.10.5",
   "author": {
     "name": "reolink"
   },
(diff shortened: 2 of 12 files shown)


reolink-net (approved):
Version sync verified against the published 0.10.5; --self reports rather than passes on a drifted manifest.

PR #35, opened 2026-07-30
installers: anchor integrity to the repository, not the release; fail closed

cuican-reolink:
Reported privately through support; a security advisory with reporter credit follows once coordinated. The finding, in one sentence: the installers verified downloads against a `SHA256SUMS` fetched from the same release as the archive, while the failure message claimed tamper protection — but whoever can replace a release asset can regenerate a matching `SHA256SUMS` in the same API call, so the check could only ever catch accidental corruption.

## What changes

**The integrity anchor moves out of the release trust boundary.** Each release now commits `checksums/<tag>.sha256` to this repository's default branch (from the machine that built the artifacts), and both installers verify against that — never against the release-attached file. The asymmetry that makes this meaningful, verified on this repository's actual settings: replacing a release asset is one API call by any account with write access and leaves no visible trace; changing a file on `main` requires a reviewed pull request and leaves permanent history.

**Fail closed, twice:**

- No fallback to the release-attached `SHA256SUMS`. A fallback would hand an attacker who controls the release a bypass: make the repository fetch fail, and the script walks back into the boundary it was escaping.
- A tag with no committed checksum file aborts. A fabricated release — a tag that never went through the release process — has no file on `main`, so it does not install.

**The wording is corrected.** A mismatch now reads "The download does not match the checksum committed to the repository", not "corrupt or tampered with". The comments state plainly that this is an integrity check, not a signature: it moves the anchor, it does not prove who built the archive.

**`check-version-sync.sh` asserts the checksum file exists** for the version under check. With fail-closed installers, a release synced without its file is not a stale badge — it is a broken `curl | sh` for every user until someone notices. The daily CI run is how someone notices.

`checksums/v0.10.5.sha256` is included for the current release, copied from the build machine's `dist/SHA256SUMS` (and cross-checked against the published assets, which still match).

## What this deliberately does not do

- **No signing (minisign/cosign) yet.** Neither verifier is preinstalled for the audience of a one-line install, which leaves two behaviours: fail without it (breaks the one-line path) or skip without it (unverified in practice plus a claim of signing — the same overstatement this PR removes, relocated). If signing lands later it layers on top of this for users who have a verifier; the two compose.
- **No GitHub attestation.** Strongest option, but requires the build to run in Actions, which it does not today. Tracked.

## Tested

Three paths, against a local server standing in for `raw.githubusercontent.com` (the tested copy differs from the shipped script only in that base URL — diff-verified):

| case | result |
|---|---|
| committed checksum matches | installs; `reolink-cli 0.10.5 (external · LAN-only)` |
| checksum altered (stands in for a replaced asset) | aborts, exit 1, zero files installed |
| no checksum file for the tag (fabricated release / unsynced) | aborts, exit 1, zero files installed |

`sh -n` passes; `./scripts/check-version-sync.sh` passes including the new existence check. `install.ps1` got the same change but no PowerShell run here — same URL, same fail-closed structure, reviewed by eye.

After merge, the real `curl | sh` path from `main` is the final acceptance run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,8 +287,14 @@ what they do:
 
 - resolve the latest release from the GitHub API, then download that release's
   asset from `github.com` — no other host is contacted
-- **verify the download against the release's `SHA256SUMS` and abort on any
-  mismatch**, missing entry, or missing checksum file
+- **verify the download against `checksums/<tag>.sha256` committed to this
+  repository's default branch**, and abort on any mismatch, missing entry, or
+  missing checksum file — never against the checksum attached to the release
+  itself. A release asset can be replaced by one API call and the attached
+  checksum regenerated with it; a file on the default branch sits behind a
+  reviewed pull request and permanent history. This is an integrity check, not
+  a signature: it moves the anchor out of the release, it does not prove who
+  built the archive.
 - install two binaries to `~/.local/bin` (`%USERPROFILE%\.local\bin` on
   Windows) — **no `sudo`, no system directories, no services**
 - stop a running `reolink-gateway` **only if it runs from that same prefix**,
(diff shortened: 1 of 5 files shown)


reolink-net (approved):
Verified the three paths locally (match installs, mismatch aborts with zero files, missing tag file aborts). The no-fallback property is the whole point — held it to that.

PR #37, opened 2026-07-30
version copies: nine locations down to seven; badge becomes a live lookup

cuican-reolink:
Follow-up to #7, attacking the other half of the problem. The tooling (`--set`, `--self`, CI) manages the copies; this PR reduces how many copies exist. A copy that cannot go stale beats a copy that is checked.

## What changes

- **README version badge → live shields.io release lookup** (`img.shields.io/github/v/release/reolink/reolink-cli`). It now always shows the latest published release, with zero maintenance. The stale-badge-over-newer-download failure that #7 documented twice becomes structurally impossible for the badge.
- **The skill's setup notes no longer name a version.** The example output reads `# → <version> (external · LAN-only)` — the line was demonstrating the flavour suffix, not the number.
- **`check-version-sync.sh` shrinks from nine locations to seven** — the seven JSON manifests, which are irreducible: plugin systems (Claude marketplace, Codex, Cursor, Gemini, openclaw, npm-style tooling) read them from the repository in place, so they must carry real numbers.
- CONTRIBUTING and script comments updated to match, keeping the history of why.

## One behaviour worth calling out

`--self` (the PR check) now effectively enforces that **a version bump and its `checksums/<tag>.sha256` travel in the same pull request** — the existence check fails otherwise. That is deliberate, not incidental: since #35 the installers fail closed on a missing checksum file, so a bump reaching `main` without its file breaks `curl | sh` for every user at that instant. The CI failure on the PR is the fence before that cliff.

## Internal counterpart (context)

The same consolidation landed upstream where the binaries are built: the six crate versions collapsed into one `[workspace.package]` inheritance, and the packaging script reads the workspace root (the member files now say `version.workspace = true`, which the old extraction would have read literally). A release bump internally is now one line plus a human-written CHANGELOG entry.

## Tested

- `sh -n` passes; the check passes against published 0.10.5; `--self` passes.
- On a scratch copy: `--set 0.11.0` rewrites exactly the seven manifests (verified by count and by re-grep), and `--self` then fails on the missing `checksums/v0.11.0.sha256` — the new fence, firing where it should.
- The dynamic badge URL was fetched and renders the current release.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,19 +27,21 @@ you want. We implement it upstream and it arrives in the next release.
   the docs that had never been executed — a flag that did not exist, a
   subcommand that had been renamed. If you touch an example, run it and paste
   the output in the PR.
-- **Keep manifests in step.** The version appears in nine places — seven
-  manifests, the README badge, and `skills/reolink-cli/references/setup.md` —
-  plus a CHANGELOG heading. They must all agree with the published release. Do
+- **Keep manifests in step.** The version appears in seven manifests, plus a
+  CHANGELOG heading and the release's `checksums/<tag>.sha256`. (It was nine:
+  the README badge is a live shields.io release lookup now, and the skill's
+  example output no longer names a version — copies that cannot go stale
+  beat copies that are checked.) They must all agree with the published release. Do
   not hand-edit a version to something no release uses.
 
   ```bash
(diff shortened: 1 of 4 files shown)


reolink-net (approved):
The seven that remain are read in place by plugin systems; the two that left can no longer go stale. The --self fence on the checksums file is the right coupling to #35.

PR #40, opened 2026-07-30
docs: make the archive example platform-agnostic and zsh-safe

### LINKED ISSUE #39: Docs: "download an archive" example hardcodes the macOS filename — breaks copy-paste on Linux (and the `cd` glob fails in zsh)
### What

In the README's "Prefer a downloadable archive?" block (line 94), the instructions say to grab the archive for your platform, but the command hardcodes the macOS (darwin-arm64) filename:

```bash
tar -xzf reolink-cli-*-external-darwin-arm64.tar.gz && cd reolink-cli-* && ./install.sh
```

A Linux user — both linux-x86_64 and linux-arm64 are listed as supported — downloads their correct archive (`…-external-linux-x86_64.tar.gz`), pastes this line, and gets:

```
tar: reolink-cli-*-external-darwin-arm64.tar.gz: Cannot open: No such file or directory
```

because the glob only matches the darwin archive. Two of the four supported platforms hit this.

### Also: the `cd` half fails on macOS's own default shell

Running the line as documented (macOS, v0.10.4 archive): after extraction the glob `reolink-cli-*` matches **both** the extracted directory and the tarball sitting next to it. zsh — the macOS default shell — treats two-argument `cd` as string substitution and fails:

```
zsh:cd:1: string not in pwd: reolink-cli-0.10.4-external-darwin-arm64
```

It only works in bash, which silently ignores `cd`'s extra arguments. So the documented line has never worked in a default macOS terminal either. A trailing slash restricts the glob to directories and works in both shells (verified in bash and zsh).

### Suggested fix

Since a user only has the one archive they just downloaded, a platform-agnostic glob works everywhere, and the trailing slash fixes the `cd`:

```diff
-tar -xzf reolink-cli-*-external-darwin-arm64.tar.gz && cd reolink-cli-* && ./install.sh
+tar -xzf reolink-cli-*.tar.gz && cd reolink-cli-*/ && ./install.sh
 # Windows: extract the .zip and run .\install.ps1
```

PR follows.

ch-bas:
## What changes

Fixes #39. The "Prefer a downloadable archive?" example hardcoded the darwin-arm64 filename in the `tar` command — a Linux user (two of the four supported platforms) pasting it got `Cannot open: No such file or directory`. Separately, the `cd reolink-cli-*` half failed in zsh, macOS's default shell: after extraction the glob matches both the extracted directory and the tarball, and zsh's two-argument `cd` does string substitution (`string not in pwd`). The line now uses a platform-agnostic tar glob (the user only has the one archive they downloaded) and a trailing slash so the `cd` glob matches directories only:

```bash
tar -xzf reolink-cli-*.tar.gz && cd reolink-cli-*/ && ./install.sh
```

## How it was verified

Ran both the old and new lines against the real v0.10.4 darwin-arm64 archive. Old line: `cd` fails in zsh with `string not in pwd` (works in bash only because bash ignores `cd`'s extra arguments). New line: extraction, `cd`, and `install.sh` lookup all succeed in **both bash and zsh**, end to end.

## Checklist

- [ ] `cargo test --workspace` passes — N/A: this repo contains no Rust source; README only

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ Grab the archive for your platform from the
 extract, and run the bundled installer:
 
 ```bash
-tar -xzf reolink-cli-*-external-darwin-arm64.tar.gz && cd reolink-cli-* && ./install.sh
+tar -xzf reolink-cli-*.tar.gz && cd reolink-cli-*/ && ./install.sh
 # Windows: extract the .zip and run .\install.ps1
 ```
 


reolink-net (approved):

PR #42, opened 2026-07-30
docs: stop telling Windows users to self-update in the quick start

### LINKED ISSUE #41: Docs: quick-start tells Windows users to `self-update`, but self-update is macOS/Linux only
### What

Near the top of the README (line 84), the upgrade instruction is unconditional:

> Already installed? Upgrade in place with `reolink-cli self-update --yes`.

Further down (under **Platform support**, ~134 lines later), the detailed section says self-update doesn't actually work on Windows:

> `self-update` covers macOS and Linux. On Windows it exits with the download link instead: the archive is a `.zip`, and a running `.exe` cannot be replaced in place — upgrade by extracting the new archive and running `install.ps1`.

So a Windows user follows the quick-start line, runs `reolink-cli self-update --yes`, gets a download link instead of an upgrade, and only finds out why ~130 lines later.

### Why it's worth fixing

It's the first upgrade instruction most people see, and it sends a third of the supported platforms (Windows x86_64 is listed as supported) down a path the tool explicitly doesn't take. A short qualifier at the top removes the surprise.

### Suggested fix

Qualify the quick-start line and point Windows users to the right step:

```diff
-Already installed? Upgrade in place with `reolink-cli self-update --yes`.
+Already installed? On macOS/Linux, upgrade in place with
+`reolink-cli self-update --yes`. On Windows, re-run the installer
+(`install.ps1`) — see [Platform support](#platform-support) for why.
```

PR follows.

ch-bas:
## What changes

Fixes #41. The first upgrade instruction in the README was unconditional — "Upgrade in place with `reolink-cli self-update --yes`" — but `self-update` covers macOS and Linux only; on Windows it exits with a download link (the archive is a `.zip` and a running `.exe` cannot be replaced in place), which the README itself only explains 134 lines later under Platform support. The quick-start line now qualifies the platforms and points Windows at re-running `install.ps1`, linking to [Platform support](#platform-support) for the reason.

## How it was verified

Cross-checked the two README statements against each other and against the v0.10.1 release notes, which describe the Windows exit-with-link behavior as intentional (`self-update` only untars, and a running `.exe` cannot be replaced in place). The `#platform-support` anchor was checked against the actual `## Platform support` heading. Docs-only; no command behavior involved.

## Checklist

- [ ] `cargo test --workspace` passes — N/A: this repo contains no Rust source; README only

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,9 @@ curl -fsSL https://raw.githubusercontent.com/reolink/reolink-cli/main/install.sh
 iwr https://raw.githubusercontent.com/reolink/reolink-cli/main/install.ps1 | iex
 ```
 
-Already installed? Upgrade in place with `reolink-cli self-update --yes`.
+Already installed? On macOS/Linux, upgrade in place with
+`reolink-cli self-update --yes`. On Windows, re-run the installer
+(`install.ps1`) — see [Platform support](#platform-support) for why.
 
 <details>
 <summary>Prefer a downloadable archive?</summary>


reolink-net (approved):

PR #47, opened 2026-07-30
fix: skill setup snippet could never download the binary

### LINKED ISSUE #43: Skill setup snippet cannot download the binary: `macos` token + missing `external` segment (same drift 0.10.1 fixed in self-update)
### What

The manual-install snippet in `skills/reolink-cli/references/setup.md` (the path an agent follows when `reolink-cli` is not on PATH) cannot download anything, for two independent reasons:

1. **Wrong OS token** (line 14): `case "$os" in Darwin) os=macos;;` — the published assets use `darwin` (`reolink-cli-0.10.5-external-darwin-arm64.tar.gz`, verified against the v0.10.5 release). The glob `reolink-cli-*-macos-arm64.tar.gz` matches nothing, so `gh release download` finds no asset. This is the same `macos`→`darwin` token drift the 0.10.1 release fixed in `self-update` ("The archives had gained an `-external` flavour segment … and the macOS token had moved from `macos` to `darwin`; neither reached the updater") — the skill snippet kept both halves of the old naming.

2. **The curl fallback drops the `external` segment** (lines 17+23): `asset="reolink-cli-*-${os}-${arch}.tar.gz"` then `name=$(echo "$asset" | sed "s/\*/${ver#v}/")` produces `reolink-cli-0.10.5-macos-arm64.tar.gz` — even with the OS token fixed, this URL 404s because the real filename has `-external-` between the version and the OS.

So on macOS both paths fail, and on Linux the curl fallback fails (the gh path happens to work there because the glob's `*` absorbs `0.10.5-external`).

### Suggested fix

`os=darwin` (and the matching Intel-guard on line 16), plus put `external` explicitly in the pattern so the sed substitution reconstructs the real filename:

```diff
-case "$os" in Darwin) os=macos;; Linux) os=linux;; esac
+case "$os" in Darwin) os=darwin;; Linux) os=linux;; esac
 case "$arch" in aarch64) arch=arm64;; x86_64|amd64) arch=x86_64;; esac
-[ "$os" = macos ] && [ "$arch" = x86_64 ] && { echo "macOS Intel not supported — Apple Silicon (arm64) only"; exit 1; }
-asset="reolink-cli-*-${os}-${arch}.tar.gz"
+[ "$os" = darwin ] && [ "$arch" = x86_64 ] && { echo "macOS Intel not supported — Apple Silicon (arm64) only"; exit 1; }
+asset="reolink-cli-*-external-${os}-${arch}.tar.gz"
```

### Related observation (maintainer's call, not included in the PR)

Unlike `install.sh` since #35, this snippet installs the downloaded binaries without any checksum verification. Now that `checksums/<tag>.sha256` is committed to the repo, the snippet could verify against it with a few lines — but that extends the trust-model decision made in #35, so flagging rather than changing it.

PR follows for the filename fixes.

ch-bas:
## What changes

Fixes #43. The skill's manual-install snippet could never download a binary: `Darwin) os=macos` while the published assets use `darwin`, and the asset pattern lacked the `external` flavour segment, so the curl fallback's sed substitution built `reolink-cli-<ver>-macos-arm64.tar.gz` — a filename that has never existed on any release. Same naming drift the 0.10.1 release fixed in `self-update`; the snippet kept the old tokens. Now `os=darwin` (and the matching Intel guard) with `reolink-cli-*-external-${os}-${arch}.tar.gz`.

## How it was verified

Ran both fixed paths against the live v0.10.5 release: `gh release download --pattern 'reolink-cli-*-external-darwin-arm64.tar.gz'` fetches the asset, and the curl-fallback filename the snippet constructs (`reolink-cli-0.10.5-external-darwin-arm64.tar.gz`) is a byte-for-byte match with an existing release asset name.

One observation left in the issue rather than this PR: the snippet installs the binaries without checksum verification, which could now use the `checksums/<tag>.sha256` model from #35 — that extends a trust-model decision, so it's flagged for a maintainer call instead of changed here.

## Checklist

### DIFF
--- a/skills/reolink-cli/references/setup.md
+++ b/skills/reolink-cli/references/setup.md
@@ -11,10 +11,10 @@ set -e
 REPO="reolink/reolink-cli"
 BIN="$HOME/.local/bin"; mkdir -p "$BIN"
 os=$(uname -s); arch=$(uname -m)
-case "$os" in Darwin) os=macos;; Linux) os=linux;; esac
+case "$os" in Darwin) os=darwin;; Linux) os=linux;; esac
 case "$arch" in aarch64) arch=arm64;; x86_64|amd64) arch=x86_64;; esac
-[ "$os" = macos ] && [ "$arch" = x86_64 ] && { echo "macOS Intel not supported — Apple Silicon (arm64) only"; exit 1; }
-asset="reolink-cli-*-${os}-${arch}.tar.gz"
+[ "$os" = darwin ] && [ "$arch" = x86_64 ] && { echo "macOS Intel not supported — Apple Silicon (arm64) only"; exit 1; }
+asset="reolink-cli-*-external-${os}-${arch}.tar.gz"
 tmp=$(mktemp -d)
 if command -v gh >/dev/null 2>&1; then
   gh release download --repo "$REPO" --pattern "$asset" --dir "$tmp"


reolink-net (approved):

PR #48, opened 2026-07-30
fix: week-table bitmap was documented backwards

### LINKED ISSUE #44: recording.md documents `--week-table` backwards — its own Mon–Sat example silently records Sun–Fri
### What

`skills/reolink-cli/references/recording.md` documents the `--week-table` bitmap in the opposite bit order from the binary, and derives wrong values from it (lines 32–35):

> `--week-table` is a **7-bit bitmap, Mon=bit0 … Sun=bit6**. Common values:
> - Weekdays (Mon–Fri): `31`
> - Mon–Sat: `63`
> - All week: `127`

The v0.10.5 binary's help for `record schedule set`:

```
--week-table <WEEK_TABLE>  Weekly day bitmap (bit0=Sun .. bit6=Sat)
```

With bit0=Sun: Mon–Fri = bits 1–5 = **62** (not 31), Mon–Sat = bits 1–6 = **126** (not 63). Only `127` is coincidentally right.

### Why it's worth fixing

This one fails silently on the camera. The file's own "Weekly window (Mon–Sat 08:00–22:00)" example (line 56) passes `--week-table 63`, which actually schedules **Sun–Fri**: recording quietly happens on Sunday and is missing on Saturday, and nobody gets an error. Documented value `31` for "weekdays" likewise records Sun–Thu.

### Suggested fix

Correct the bit order and the derived values (`31`→`62`, `63`→`126`, keep `127`), including the example.

PR follows.

ch-bas:
## What changes

Fixes #44. `recording.md` documented `--week-table` as "Mon=bit0 … Sun=bit6" with weekdays=`31` and Mon–Sat=`63`; the binary's help says `bit0=Sun .. bit6=Sat`, so those values schedule the wrong days — `31` is Sun–Thu, and the file's own "Weekly window (Mon–Sat)" example passed `63`, which is Sun–Fri: recording silently happens on Sunday and is missing on Saturday, with no error anywhere. Corrected to the binary's bit order with weekdays=`62`, Mon–Sat=`126` (all-week `127` was already right), example updated to `126`.

## How it was verified

Bit order taken from the v0.10.5 binary's `record schedule set --help` ("Weekly day bitmap (bit0=Sun .. bit6=Sat)"); values re-derived from that order. Stated plainly: no camera write was performed to verify device-side behaviour — the fix trusts the binary's own help over the doc's contradicting claim.

## Checklist

### DIFF
--- a/skills/reolink-cli/references/recording.md
+++ b/skills/reolink-cli/references/recording.md
@@ -29,9 +29,9 @@ reolink-cli --camera front-door --gateway-addr 127.0.0.1:9000 record config get
 
 ## Rules
 
-- `--week-table` is a **7-bit bitmap, Mon=bit0 … Sun=bit6**. Common values:
-  - Weekdays (Mon–Fri): `31`
-  - Mon–Sat: `63`
+- `--week-table` is a **7-bit bitmap, Sun=bit0 … Sat=bit6** (matches `--help`). Common values:
+  - Weekdays (Mon–Fri): `62`
+  - Mon–Sat: `126`
   - All week: `127`
 - **Must** read current (`record schedule get`) before overwriting — SET replaces the full schedule object.
 - **Must** warn before `record schedule set --disable` — no new recordings will be written until re-enabled.
@@ -53,7 +53,7 @@ reolink-cli --camera front-door record schedule set --disable
 ## Weekly window (Mon–Sat 08:00–22:00)
 
 ```bash
-reolink-cli --camera front-door record schedule set --enable --plan-type weekly --week-table 63 \
+reolink-cli --camera front-door record schedule set --enable --plan-type weekly --week-table 126 \
   --start-hour 8 --start-min 0 --end-hour 22 --end-min 0
 ```
 


reolink-net:
Holding this one, and I want to be straight about why rather than sit on it silently.

You are right that something is wrong: the doc says `Mon=bit0 … Sun=bit6` and `record schedule set --help` says `bit0=Sun .. bit6=Sat`. They contradict each other, so one of them has been wrong the whole time.

But "matches `--help`" is a consistency argument, not a correctness one. If the help string is the wrong one, this PR makes both places agree on the wrong bit order — and a reader would then have two sources confirming each other, which is worse than the current visible contradiction.

I tried to settle it on hardware and could not:

- The v20 protocol header defines the field as a bare `int week_table` with no comment on bit order.
- `record schedule` is cmd 594/595, which is **battery-device-only** — my mains camera answers `unsupported: record.schedule (cmd 594/595) is battery-device-only; this firmware doesn't expose the long-run-mode config endpoint`.
- The four battery devices I have access to were all unreachable when I tried.

So I have no way to observe which day a bit actually selects.

**Do you have a battery device you can set a schedule on?** If so, one experiment settles it permanently: set `--week-table 1` (bit0 only), then look at which day is checked in the app or the device's own web UI. Post what you see and I will merge with the verified order and a note saying it was checked against a device.

If you do not, here is what I have done in the meantime, on the main branch upstream — the doc now says:

> `--week-table` is a 7-bit day bitmap. **Read the bit order from `record schedule set --help`** — do not compute it from memory; this document previously carried a bit order that contradicted the CLI's own help, and neither had been verified against a device.

and the weekly example switches to `--week-table 127` (all week), which is the one value that means the same thing under either bit order.

That is deliberately less useful than your table. It is also the only thing I can write that I know to be true. Once someone verifies the order on a device, the table comes back with the numbers filled in.

Your finding stands regardless — the contradiction was real and worth reporting. This is about which half gets corrected, not whether there was a bug.

reolink-net (approved):
Verified in the vendor SDK after my earlier comment — you are right. Details in the follow-up comment.

reolink-net:
Merged. Retracting my previous comment: I asked you to run an experiment I should have been able to settle myself, and the answer was sitting in the vendor SDK the whole time. I had grepped the wrong header.

`BCNetSDK.h`, in the weekly-plan structure itself:

```c
/**
 * @param iWeekTable
 * index 0:sunday
 * index 1~index 6:Monday->Saturday
 */
int iWeekTable[7];
```

Four more places in the same header agree, including the bitmap form of the field:

```
unsigned int weekdays_bitmap;  // bit0 sunday bit1: ~ bit6:
int iStartWeekday;             // 0: sunday
int iWeek[7];                  // 0=Sunday ... 6=Saturday
E_BC_RA_EVERY_SUNDAY, E_BC_RA_EVERY_MONDAY, …
```

So `bit0 = Sun` is correct, the CLI's `--help` was right, and the doc had been wrong since it was written. Your numbers check out too — with Sun at bit 0, Mon–Fri is bits 1–5 = 62 and Mon–Sat is bits 1–6 = 126, which is exactly what you filed. The old 31 and 63 are what you get by computing the same days under the wrong bit order, which is a good sign the doc was written from the stated (wrong) premise rather than from a device.

One correction to how I framed the merge criterion, since it affected you: I said "matches `--help`" was a consistency argument rather than a correctness one, and held the PR on that basis. That reasoning was sound — but the right response to it was to go find the authority, not to hand the verification back to you. Sorry for the round trip.

The upstream doc now carries your table with the source cited, so the next person who wonders can check it without owning a battery device.

PR #50, opened 2026-07-30
docs: three small drifts — deprecated flag example, missing endpoint, stale count

### LINKED ISSUE #46: Three small doc drifts: deprecated `--duration` example, missing `/api/preview/video`, stale "nine" in workflow comments
Three small drifts, none broken-if-copied, grouped to keep the tracker tidy:

1. **`references/controls.md:17`** — the spotlight example uses `--duration 30`, which the v0.10.5 binary marks deprecated in its own help: *"DEPRECATED: re-arming extends the on-period, it does not pulse — use `light spotlight blink` for repeated flashes"*. The doc should carry the same warning (the README even advertises "native blink" as a highlight).

2. **`references/gateway-http.md`** — SKILL.md (line 263) lists four streaming endpoints and calls this file "the full spec", but the Streaming Endpoints section documents only three: `/api/events`, `/api/snapshot`, `/api/vod/download`. `/api/preview/video` is missing. (Its exact query parameters beyond `token` don't appear to be documented anywhere in the repo, so the addition can only state what's verifiable.)

3. **`.github/workflows/version-sync.yml:1,13`** — the header comments still say "nine places" / "do the nine agree", but #37 cut the list to seven; the script's own header (`check-version-sync.sh:11`) already says "seven manifests (it was nine …)". Comment-only drift.

(Deliberately not included: `CHANGELOG.md:97` also says "nine", but that's the 0.10.4 entry describing the state when 0.10.4 shipped — a historical snapshot, accurate at the time.)

PR follows.

ch-bas:
## What changes

Fixes #46. Three small drifts, one commit:

- **controls.md** — the spotlight example used `--duration 30`, which the v0.10.5 binary marks deprecated in its own help ("re-arming extends the on-period, it does not pulse — use `light spotlight blink`"). The example now omits it and carries the same warning.
- **gateway-http.md** — SKILL.md lists four streaming endpoints and calls this file "the full spec", but it documented three. `/api/preview/video` is now named with the same token-in-query rule. Only what's verifiable is stated — its further query parameters are documented nowhere in the repo, so no example URL is invented.
- **.github/workflows/version-sync.yml** — header comments still said "nine places" / "do the nine agree" after #37 cut the list to seven (the script's own header already says seven). Comment-only; no workflow logic touched.

## How it was verified

Deprecation text and endpoint list cross-checked against the v0.10.5 binary's `light spotlight set --help` and SKILL.md line 263; the workflow change is comments only, and `--self` still passes locally (`./scripts/check-version-sync.sh --self`).

## Checklist

- [ ] `cargo test --workspace` passes — N/A: this repo contains no Rust source; docs + workflow comments only

### DIFF
--- a/.github/workflows/version-sync.yml
+++ b/.github/workflows/version-sync.yml
@@ -1,4 +1,4 @@
-# The version appears in nine places in this repository plus a CHANGELOG
+# The version appears in seven places in this repository plus a CHANGELOG
 # heading, and keeping them in step by hand has failed three times — once
 # leaving the plugin manifests behind, once leaving the whole repository a
 # version back while the release was already published (readers saw a 0.10.0
@@ -10,7 +10,7 @@
 #
 # Two different questions, so two different triggers:
 #
-#   --self   do the nine agree with each other?  → every PR and push.
+#   --self   do the seven agree with each other?  → every PR and push.
 #            A branch may legitimately carry a version no release uses yet, so
 #            this deliberately does not look at the published release.
 #
--- a/skills/reolink-cli/references/controls.md
+++ b/skills/reolink-cli/references/controls.md
@@ -13,8 +13,9 @@ reolink-cli --camera front-door light ir set --state auto    # or on / off
 reolink-cli --camera front-door light statusled get
 reolink-cli --camera front-door light statusled set --state off
 
-# Spotlight (manual, 0 = until turned off)
-reolink-cli --camera front-door light spotlight set --enable --duration 30
+# Spotlight (manual; --duration is deprecated — re-arming extends the
+# on-period, it does not pulse; use `light spotlight blink` for flashes)
+reolink-cli --camera front-door light spotlight set --enable
 reolink-cli --camera front-door light spotlight set --disable
 
 # White LED alarm config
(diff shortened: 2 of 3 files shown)


reolink-net (approved):
All three verified: the workflow comment still said nine (my own PR #37 updated the script and CONTRIBUTING but missed the YAML), /api/preview/video exists in web/mod.rs, and --duration is marked DEPRECATED in the arg help.

reolink-net:
Owing you a correction: this PR was reverted about an hour after it merged, by me, and I only noticed because I went looking for the next batch of defects rather than because anything flagged it.

The `release: 0.10.6` sync copies the skill docs from an internal repository into this one. Those internal copies had never received your changes, so the copy silently deleted:

- the `/api/preview/video` note you added to `gateway-http.md`;
- the conflict resolution on the spotlight example, which put `--duration 30` back into a command whose own comment calls the flag deprecated.

Neither review nor CI caught it, and the person doing it — me — did not either. Both are restored in #53, along with the thing that should have existed first: `scripts/check-docs-mirror.sh`, which compares the two copies and specifically names the dangerous direction, lines that exist only here and would be deleted by a sync. The release process now requires running it before the copy is made.

The underlying problem is that these docs exist twice with no mechanism keeping them in step, which is the same shape as everything else you have been finding: one fact, two copies, and the copy drifts. Version numbers were in sixteen places (now two), `--help` text was restated in prose (now cited), and the docs themselves are duplicated across repositories. The checker makes that last one loud; the actual fix is one copy, which is a bigger change than I have made yet.

Two other things fell out of the same sweep, both in the class you have been mining:

- **the gateway endpoint list was a third complete** — 14 endpoints exist, 3 were documented, which is why `/api/preview/video` was missing for you to find in the first place;
- **the "needs the gateway" list was wrong**: `features`, `doctor` and `cache status|clean` all work without it, verified by running each with nothing listening.

Sorry for undoing your work. It was merged, then quietly discarded, and you would have had every reason to be annoyed if you had found it before I did.

PR #51, opened 2026-07-31
docs: cite the source for values; check documented commands mechanically

cuican-reolink:
Follow-up to #47 through #50. Those six PRs were all the same defect wearing different clothes: a fact that lives in the code, restated in prose, where the restatement drifted and nothing noticed. `--help` said 1–50 and the doc said 0–100; `--help` said DEPRECATED and the doc recommended the flag; `install.sh` said `darwin` and the skill snippet said `macos`. This PR goes after the two halves of that.

## Cite the source, or stop restating it

**`--week-table` now names its authority** rather than pointing at `--help`. #48 was right, and I can now say *why* rather than "because it matches the help text": the vendor SDK defines the field in the weekly-plan struct itself —

```c
/**
 * @param iWeekTable
 * index 0:sunday
 * index 1~index 6:Monday->Saturday
 */
int iWeekTable[7];
```

— with four corroborating definitions in the same header (`weekdays_bitmap // bit0 sunday`, `iStartWeekday // 0: sunday`, `iWeek[7] // 0=Sunday ... 6=Saturday`, and the `E_BC_RA_EVERY_SUNDAY, E_BC_RA_EVERY_MONDAY` enum order). The note also records that the doc carried Mon=bit0 until today, because a value copied from it before then selects the wrong days and nothing about the resulting schedule looks broken.

**`--sensitivity` and spotlight `--duration` go the other way**: the doc no longer restates the accepted range or the deprecation status, it points at `--help`, which is generated from the code and cannot drift. What stays is the part `--help` cannot say — that motion and AI use *different* scales, which is the trap, not the numbers.

The rule this follows: state a fact in prose only when prose is the only place it can live. Otherwise cite where it lives.

## Check the documented commands mechanically

`scripts/check-doc-commands.py` extracts every `reolink-cli …` line from fenced code blocks and feeds it to the real binary's parser. clap validates subcommands, flag names, and value ranges during parsing, before a command does anything — so those three layers are checkable, and this checks them.

```
$ ./scripts/check-doc-commands.py --cli reolink-cli
checked 169 documented commands (31 skipped as placeholders)
all of them parse.
```

Safety: commands run against TEST-NET-1 (192.0.2.0/24, unroutable per RFC 5737) with the gateway pointed at a closed port and config paths redirected to a temp dir, so anything that parses cleanly dies at the first connection attempt without touching a device or the real config.

Verified it catches rather than just passes: restoring `--sensitivity 60` (the #49 defect) makes it report

```
invalid value '60' for '--sensitivity <SENSITIVITY>': 60 is not in 1..=50
```

Getting to a clean run took three rounds of fixing the checker, not the docs — it first reported 121 findings because it compared subcommand flags against the root help, then 31 because it read trailing prose (`(recent fires)`) as arguments. Both were the checker misreading, and the current version only scans fenced blocks and skips PowerShell wrappers for that reason.

**What it cannot see**, stated plainly because the gap matters: whether a documented *meaning* is correct. A bit order, a deprecation note, a claim about Windows — all parse fine and can all be wrong. That class is what a human reader catches, and reducing it is the job of the first half of this PR, not the second.

### DIFF
--- a/scripts/check-doc-commands.py
+++ b/scripts/check-doc-commands.py
@@ -1,0 +1,141 @@
+#!/usr/bin/env python3
+"""Run every `reolink-cli …` command that appears in the docs through the real
+binary's argument parser, and report the ones it rejects.
+
+    ./scripts/check-doc-commands.py [--cli path/to/reolink-cli] [docs-root ...]
+
+Why this exists: the reference docs assert what the CLI accepts, and nothing
+verified those assertions. An outside contributor read them against the binary
+and filed six defects in one day — one of which (`detect motion set
+--sensitivity 60`, valid range 1–50) this script now catches on its own. Code
+has a compiler and a test suite; prose had neither.
+
+What it can and cannot see. clap validates subcommands, flag names, and value
+ranges during parsing, before any command runs — so those are checkable, and
+this script checks them. It cannot see whether a documented *meaning* is right:
+a bit order, a deprecation note, a claim about Windows. Those need a reader.
+Reducing that second class is a docs-design job (point at `--help` instead of
+restating it), not something a checker can do.
+
+Safety: commands run against TEST-NET-1 (192.0.2.0/24, unroutable by RFC 5737)
+with the gateway pointed at a closed port and config/registry paths redirected
+into a temp dir, so a command that parses cleanly fails at the first connection
+attempt without touching a device or the real config.
+"""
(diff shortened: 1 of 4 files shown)


reolink-net (approved):
The week-table note now carries its authority; the checker's limits are stated where a reader will see them.

Issue #36, opened 2026-07-30
Difference in output between Windows and Linux "discover" command


ian1182:
### Version

reolink-cli 0.10.5 (external · LAN-only)

### Platform

Linux arm64

### Camera model and firmware

Not applicable

### Command you ran

```shell
reolink-cli discover
```

### What you expected, and what happened instead

There seems to be a difference between the output produced by the windows and linux_arm64 versions of the reolink-cli command. Both provide useful fields but the UID value is missing from some of the linux output. In the examples I have only included 2 devices to show the differences. Firstly a RLN12W and secondly a plug-in wifi doorbell (i.e. not battery).

The linux output seems a bit more comprehensive for some devices but lacks the UID value for them (RLN12W in this example) but the wifi doorbell does. Also the scopes field sometime contains the device name and other time none. It would be useful for the linux version to always include both the UID and device name in scopes as that helps reconcile which device it is.

What I did see is the "counts" field in the output between Windows and Linux have different values for "lan" and "lanUdp".  The total number of devices was 11 in each case but it seems windows classified all as "lanUdp" but Linux had 4 under "lan" and 7 "lanUdp".

The device which appeared under "lan" were a RLN12W, HomeHub (BASE_WENNT6NA5), OMVI 3i POE and E1 Outdoor (IPC_566SD85MP). And devices in "lanUdp" included a doorbell (DB_566128M5MP_W), CX810 and Lumus (IPC_FH1NA48MPC7 & IPC_NT1NO24MP). I have not checked the network parameters for each device yet but will do so.

This is less about the "lan" vs "lanUdp" issue as such but rather the UID and device name missing for devices found under "lan". Those I believe are very useful fields in uniquely identifying a device.

In the included files I have redacted the "UID" and "urn:uuid" plus masked out the device names

### Output

```shell
Windows

{
    "ok": true,
    "command": "discover",
    "protocol": null,
    "data": {
        "source": "lan+remoteUid+lanUdp+v30Lan",
        "timeoutSecs": 2,
        "count": 11,
        "counts": {
            "lan": 0,
            "lanUdp": 11,
            "remoteUid": 0,
            "v30Lan": 0
        },
        "devices": [
            {
                "discovery": "lanUdp",
                "transport": "tcp",
                "host": "192.168.1.30:9000",
                "uid": "952700#redacted#",
                "scopes": [
                    "reolink-lan://name/NVR-XXXXXX"
                ]
            },
            
            {
                "discovery": "lanUdp",
[...]

PR #52, opened 2026-07-31
release: 0.10.6

cuican-reolink:
Syncs the repository to 0.10.6 and lands `checksums/v0.10.6.sha256` **before** the release is published — the installers resolve the latest release and then look for its checksum file on `main`, so publishing first would fail-close every `curl | sh` until this merged.

Includes the six documentation fixes from @ch-bas (#40, #42, #47, #48, #49, #50) and the `discover` merge fix from #36.

`check-version-sync.sh` and `check-doc-commands.py` both pass.


reolink-net (approved):
Checksums first, release second — verified the ordering reasoning holds.

PR #53, opened 2026-07-31
docs: restore a reverted contribution, complete the endpoint list, check the mirror

cuican-reolink:
A sweep of the defect classes @ch-bas has been finding, run against the code rather than waiting for the next report. It found four things, and the first one is ours to apologise for.

## We reverted a merged contribution within the hour

The `release: 0.10.6` sync copied this project's internal copies of `gateway-http.md` and `controls.md` over the public ones. The internal copies had never received #50, so the sync silently deleted:

- the `/api/preview/video` endpoint note that #50 added;
- the conflict resolution on the spotlight example — `--duration 30` came back into a command whose own comment calls the flag deprecated.

Nothing caught it. Not review, not CI, not the person running the sync. Both are restored here.

**Root cause: the skill docs live in two repositories and are mirrored by hand.** `scripts/check-docs-mirror.sh` now compares them and, importantly, names the dangerous direction — lines that exist *only* in the public copy, which a sync would delete. The release guide requires running it before the sync copy, which is the moment work gets destroyed. The real fix is one copy instead of two; this makes the divergence loud until then.

## Three genuine defects the sweep turned up

**The gateway endpoint list was a third complete.** The gateway serves 14 `/api/*` endpoints; `gateway-http.md` documented 3. Now all of them are listed, with the dashboard-facing ones in their own table, and the doc states that this is the complete set and asks to be told if it is not — a list that does not claim completeness is harmless, one that claims it falsely misleads.

**The "needs the gateway" list was wrong in the direction that costs the most.** SKILL.md said only `device *`, `config init` and `discover` work without a gateway. Verified by running each command with nothing listening: `features`, `doctor` and `cache status|clean` work fine too — and those are exactly what an agent reaches for first, so they were being gated behind a daemon they never needed.

**A stale example carried a deprecated flag** — covered above with the sync revert.

## What the sweep checked and found clean

Stated because "we looked" is worth as much as "we found":

- **74 shell blocks** across the docs: no syntax errors.
- **170 documented commands**: all parse against the real binary (`check-doc-commands.py`).
- **`cache clean` is dry-run by default** — verified by pointing it at a temp cache with three expired files and confirming all three survived a run without `--apply`.
- **The interactive password prompt behaves as documented**: a TTY prompts; a non-TTY fails with a message naming `--password-stdin` rather than silently storing nothing.
- **`--password` guidance is consistent** across README, AGENTS.md, SKILL.md and the references — all say do not put it on argv.
- **Version strings** agree across all seven manifests and the checksum file.

## The pattern, since it keeps repeating

Every defect in this sweep is the same shape as the six in #40–#50: one fact, two copies, and the copy drifted. Version numbers (sixteen places, now two), `--help` text restated in prose (now cited instead), and now the docs themselves existing twice. The checkers help, but each one only makes a specific drift loud — the thing that actually removes a class is deleting the second copy.

### DIFF
--- a/scripts/check-docs-mirror.sh
+++ b/scripts/check-docs-mirror.sh
@@ -1,0 +1,63 @@
+#!/bin/sh
+# The skill docs exist in two repositories and are mirrored by hand. This checks
+# they still agree.
+#
+#   ./scripts/check-docs-mirror.sh [path-to-public-clone]
+#
+# Why this exists: on 2026-07-31 a contributor's merged pull request was undone
+# within the hour. It had landed in the public repository; the release sync then
+# copied this repository's older copy over the top, and nothing noticed — not
+# review, not CI, not the person doing it. Two files, one fact, no mechanism.
+#
+# The real fix is one copy, not two. Until the mirror goes away, this at least
+# makes divergence loud, and it belongs in the release checklist BEFORE the
+# sync copy is made — that is the moment work gets destroyed.
+#
+# A difference is not automatically wrong: content for an unreleased change
+# legitimately sits here first. The point is that you look at it and decide,
+# rather than copying over the top and finding out later.
+set -eu
+
+here=$(cd "$(dirname "$0")/.." && pwd)
+public=${1:-$here/../reolink-cli-github}
+
+[ -d "$public/skills/reolink-cli" ] || {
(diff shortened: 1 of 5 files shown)


reolink-net (approved):
The revert was ours and the mirror check is the right guard for it.

Issue #54, opened 2026-07-31
Please publish an aarch64-unknown-linux-musl build (would run on Home Assistant OS)


Pharkie:
### What are you trying to do?

Run `reolink-cli` on a **Home Assistant Green** (Home Assistant OS), so Home
Assistant can call `audio talk` to push a spoken reply to a camera speaker —
"coming now", "leave it by the door" — from a wall panel.

`reolink-cli` is the only route to this. Home Assistant's own Reolink
integration exposes exactly two services, `play_chime` (a preset tone) and
`ptz_move`; there is no two-way audio. `reolink_aio` has `play_quick_reply()`,
but my cameras sit behind an NVR and reject it (`device rejected command 427`),
so streaming PCM with `audio talk` is the only thing that works — and it works
very well. From macOS I have pushed TTS clips to a camera and heard them come
out of the speaker, `ok:true`, 95,225 samples for a 6-second clip.

**The only blocker is the binary.** Home Assistant OS is Alpine/musl, and the
published `linux-arm64` asset is glibc:

```
$ file reolink-cli-0.10.6-external-linux-arm64/bin/reolink-cli
ELF 64-bit LSB pie executable, ARM aarch64, version 1 (SYSV), dynamically
linked, interpreter /lib/ld-linux-aarch64.so.1, for GNU/Linux 2.0.0, stripped

$ ./reolink-cli --version          # on Home Assistant OS
Error relocating ./reolink-cli: __res_init: symbol not found
```

Both the HAOS host **and** the `homeassistant` core container — which is where
`shell_command` actually runs, and therefore the one that matters — are aarch64
Alpine with `/lib/ld-musl-aarch64.so.1`.

So: **please could you add `aarch64-unknown-linux-musl` to the release
matrix?** You already publish four targets, so I am hoping this is close to a
one-line CI addition. It would cover Home Assistant OS with no extra runtime,
and Alpine and most slim Docker images along with it. A statically linked musl
build might even be a reasonable replacement for the current glibc
`linux-arm64` asset rather than an addition, since it would run in both places.

### How do you do it today?

From a Mac on the same LAN, which works but cannot be automated — a doorbell
reply that only fires when somebody happens to have a laptop awake is not
really a feature.

Alternatives I costed and would rather avoid:

- Running the glibc binary in a glibc sidecar container on the Home Assistant
  box. Works, but it is another container to own and roughly 80 MB of image on
  a 28 GB eMMC that has filled up once already.
[...]

PR #56, opened 2026-08-03
Installer integrity fixes (GHSA-65x2-w384-qp7j), musl builds (#54), and doc drift

cuican-reolink:
Carries the fixes for a privately reported set of installer findings, the musl support requested in #54, and three doc drifts found sweeping for the same class.

**Deliberately not merged yet.** `install.sh` is served from `main`, so merging makes it live immediately, and the README would claim musl archives are published before any release carries them. This merges as part of the next release.

## Security

Reported privately as GHSA-65x2-w384-qp7j. Two findings valid as filed, one half right, and checking them turned up a worse hole in a path the report did not cover.

**`self-update` verified nothing at all.** It resolved the latest release, downloaded the archive, unpacked it and overwrote both binaries — no checksum, no signature — with its own `REOLINK_UPDATE_REPO` override and nothing anchoring it. The installers got a committed-checksum anchor in 0.10.5; this path did not, and it is the one that runs on machines that already have the tool, repeatedly, often with `--yes`. (Fix is in the private repo; this PR carries the installer half.)

**The checksum source followed `REOLINK_REPO`.** Archive *and* checksum were both fetched from the overridden repository, so pointing it elsewhere meant the archive was validated against that repository's own committed checksums — attacker supplies both halves, every check passes, installer prints `ok`. Now pinned to `reolink/reolink-cli` in all three paths. The download stays overridable, so a mirror serving identical bytes works and a fork serving its own build does not:

```
REOLINK_REPO=attacker/evil
  archive  <- https://github.com/attacker/evil/releases/download/v0.10.6/…
  checksum <- https://raw.githubusercontent.com/reolink/reolink-cli/main/checksums/v0.10.6.sha256
```

**A crafted archive could choose which binary got installed.** Binaries were located with `find "$tmp" -name reolink-cli | head -n1` — the whole extraction tree, first hit in walk order. An archive carrying a second `reolink-cli` under an earlier-sorting directory won, and that file was then made executable and run. Reproduced before fixing:

| | archive with a planted `aaa-evil/bin/reolink-cli` |
|---|---|
| before | installs `PLANTED` |
| after | installs `REAL` |

All paths now derive the directory name from the asset name and read `bin/<name>` from it, refusing a symlink or reparse point there.

The other half of that finding — `..`/absolute members escaping the extraction directory — **did not reproduce**. BSD tar refuses `..` and exits non-zero; GNU/busybox tar strips the prefix and keeps the file inside the destination. A pre-scan refusing both member types was added anyway, so the guarantee comes from this script rather than from whichever tar is installed.

**The README pointed manual verifiers at the anchor it had just discredited** — two passages sent readers to the release-attached `SHA256SUMS` while a third explained why that file cannot be trusted. Now consistent, with a worked example.

**`SECURITY.md` states what verification does and does not prove**, as a table of covered and uncovered threats. Uncovered: a compromised maintainer account and a compromised build machine. The checksum is written by the same pipeline that builds the archive, so integrity here is not authenticity. Closing that needs a signature anchored outside the pipeline; GitHub build attestation is not currently possible because releases are not built in Actions — one workspace dependency lives outside the repository.

`install.ps1` parses cleanly under real pwsh 7.4.2, and the new zip member pre-scan was executed against hostile archives (both `../` and absolute members refused). The `Expand-Archive` copy path could not be exercised on Windows.

## musl (#54)

Alpine and what is built on it — Home Assistant OS raised this — have no glibc, so the existing archives do not run badly there, they cannot load at all. Verified on containers:

| archive | Alpine arm64 | Debian arm64 |
|---|---|---|
| `linux-arm64` (glibc) | ✗ `not found` — the missing-interpreter failure behind the reporter's `__res_init` | ✓ |
| `linux-arm64-musl` (new) | ✓ | — |
| `linux-x86_64-musl` (new) | ✓ (amd64 Alpine) | — |

Detection reads the loader path, with `ldd --version` as a second probe — on Alpine that prints "musl libc" while exiting non-zero, so it is read for output, not status. x86_64 is included although only arm64 was asked for: once the installer detects musl, shipping one architecture means x86_64 Alpine detects musl and finds no archive, which is worse than not detecting it.

These are additions — the glibc archives keep their names, because an install reconstructs the filename it came from when it self-updates.

## Doc drift

The skill's install snippet is now a call to `install.sh`. It was the fourth copy of platform detection, had already drifted once (#43), and verified nothing it downloaded.

`AGENTS.md`/`GEMINI.md` still described a private vendor tarball — "get a fresh tarball from their vendor", "no in-place network upgrade" — while line 131 of the same file documented `self-update`. The troubleshooting table listed an error string present in neither installer. `GEMINI.md` also carried the stale "works without the gateway" list corrected in `SKILL.md` a release earlier.

`scripts/check-docs-mirror.sh` is removed here: it looks for `plugins/reolink-cli/`, which exists only in the internal repository, so it has never run in this one.


reolink-net (approved):
Reviewed against the code and the reported findings. The checksum-source pinning, the derived-path extraction, and the README/SECURITY.md corrections match what was verified: the fork-override path now fetches its checksum from the canonical repository, the planted-binary archive installs the real binary, and the tar pre-scan refuses both hostile member types. Approving for release.

PR #57, opened 2026-08-03
release: 0.10.7

cuican-reolink:
Version bump, CHANGELOG, and the six artifact checksums.

The checksums must land on the default branch **before** the release is published: both installers and `self-update` now fail closed when the committed checksum file is missing, so publishing first would break every one-line install until this merged.


reolink-net (approved):
Version strings agree across all seven manifests, the CHANGELOG entry is dated, and the six checksums match the built artifacts. Checksums land before the release, as required by the fail-closed installers.

PR #58, opened 2026-08-03
docs: Home Assistant OS install path, and what to do when verification refuses

cuican-reolink:
Written after verifying the published v0.10.7 installer on Alpine arm64, Alpine amd64 and Debian arm64 — all three resolve the correct archive and verify against the pinned checksum. Two gaps showed up that the release notes did not close.

**The musl build has no user-visible marker.** `--version` prints the same string for both C libraries, so someone on Home Assistant OS cannot tell which archive they are running, and moving a binary between hosts fails with `__res_init: symbol not found` — an error naming nothing relevant. `file "$(command -v reolink-cli)"` is now documented as the way to tell.

**Home Assistant OS needed a specific answer**, since it is the case that prompted the musl build (#54): the install has to happen inside the `homeassistant` core container, because that is where `shell_command` runs and it is aarch64 Alpine. Installing on the host produces a working binary the caller cannot see.

**Both installers said what went wrong on a checksum mismatch and nothing about what to do next.** A failed integrity check is exactly when someone starts looking for a way around it, and the two obvious ways — pointing `REOLINK_REPO` elsewhere, or fetching the archive by hand — both skip the check that just fired. The message now says retry once, then report, and states plainly that there is no flag to skip verification. Same guidance in SECURITY.md, the agent guides and the skill reference, including that a missing-checksum failure is our release process being wrong rather than an attack.

`install.sh` re-verified end to end on Alpine after the edit; `install.ps1` re-parsed under pwsh 7.4.2.


reolink-net (approved):
Verified against the three real installs this came out of. The Home Assistant OS guidance matches where shell_command actually runs, and the mismatch message now points at the one correct next step instead of leaving the reader to invent a workaround.

PR #59, opened 2026-08-03
build: verify the committed checksums against what the release actually serves

cuican-reolink:
The committed checksum file became the trust anchor for every download path in 0.10.7 — both installers and `self-update` verify against it and fail closed — and nothing checked that it describes the archives being served. The existing check only asserted the file exists.

**Present but wrong is worse than missing.** Missing fails with a message naming the cause; wrong fails with `checksum mismatch`, which is the wording reserved for a tampered download, so the first thing a broken release does is send users looking for an attacker. It would also break every one-line install at once, since there is deliberately no way to skip verification.

**No download needed.** GitHub computes a digest per release asset and serves it in the API, so this compares our committed hash against the hash the server reports for the bytes it hands out. Six assets, one API call.

Verified by fault injection rather than by reading it:

| injected fault | reported |
|---|---|
| one hash corrupted | `committed deadbeef0fb6… vs served 323c84390fb6…` |
| entry for an asset not on the release | `not attached to the release` |

Both fail the run; restoring the file returns it to green.

Skipped for `--self` (a branch may legitimately name a release that does not exist yet) and when `gh` or the network is unavailable — but **the skip now prints a line**. Quietly skipping is how a check stops running without anyone deciding it should: the run stays green and looks identical to one that verified everything. The workflow gets `GH_TOKEN` for the same reason — without it the call fails, the script skips, and the job goes green having checked nothing.

### DIFF
--- a/scripts/check-version-sync.sh
+++ b/scripts/check-version-sync.sh
@@ -161,6 +161,78 @@ else
   fail=1
 fi
 
+# Existing is not the same as correct. That file is the trust anchor every
+# download path now verifies against, and nothing checked that it describes the
+# archives actually being served. A hash that is present but wrong is worse than
+# one that is missing: missing fails with a message naming the cause, wrong fails
+# with "checksum mismatch", which is the wording reserved for a tampered
+# download — so the first thing it does is send users looking for an attacker.
+#
+# No download needed: GitHub computes a digest per asset and serves it in the
+# release API, so this compares our committed hash against the one the server
+# reports for the bytes it is handing out.
+#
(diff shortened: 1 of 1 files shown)


reolink-net (approved):
Fault-injected both failure modes before approving; the skip-visibility line is the right call given the check is only reachable with a token.

reolink-net:
Re-reviewed after the rewrite: the workflow change is dropped and the API call is anonymous `curl` + `python3` instead of `gh`, so no token and no `workflow` scope are needed. Fault injection re-run against this version — corrupted hash and phantom asset both fail the run, restoring the file returns it to green.

PR #61, opened 2026-08-03
fix(windows): installer broken on Windows PowerShell 5.1, and the one-liner fails in cmd.exe

cuican-reolink:
Both found by a user pasting the documented command into the Command Prompt.

**The installer was broken on Windows PowerShell 5.1.** The archive pre-scan added for GHSA-65x2-w384-qp7j finding 3 calls `[System.IO.Compression.ZipFile]`. Windows PowerShell 5.1 — `powershell.exe`, the default on Windows — does not load that assembly until something asks for it, while PowerShell 7 has it already. The hardening meant to make the installer safer instead made it abort before touching anything, on the shell most users are running. `Add-Type` now loads it explicitly.

This is precisely the risk the previous change called out as unverified: `install.ps1` was parse-checked but never executed on Windows, and parsing was never going to catch a type that resolves at runtime.

**The one-liner failed in cmd.exe.** `iwr` is a PowerShell alias, so pasting the documented command into the Command Prompt gives:

```
'iwr' 不是内部或外部命令，也不是可运行的程序
```

an error naming the alias rather than the cause, on a machine where the user did nothing wrong. The README labelled the block "Windows (PowerShell)", but that label sits inside the code fence and is easy to miss when the prompt already says `C:\Windows\System32`. Every doc now shows `powershell -NoProfile -Command "..."`, which works from either shell, with the reason stated rather than left as a label.

**The skill's setup reference still sent Windows users to download the .zip by hand**, skipping the checksum verification every other path enforces — the same gap that removing the shell snippet was supposed to close.

Verified on PowerShell 7 (parse clean, `Add-Type` harmless, type resolves). **Not yet verified on Windows PowerShell 5.1** — that needs a Windows box.


reolink-net (approved):
The 5.1 assembly-loading gap is a real regression from the hardening change and the fix is the standard one. Merging so main stops serving a broken installer; runtime verification on 5.1 still pending.

Issue #60, opened 2026-08-03
0.10.7 Install fails on Windows 11


ian1182:
### Version

0.10.7

### Platform

Windows x86_64

### Camera model and firmware

N/A

### Command you ran

```shell
iwr https://raw.githubusercontent.com/reolink/reolink-cli/main/install.ps1 | iex
```

### What you expected, and what happened instead

Successful installation of 0.10.7. The "iwr" command had worked for updates prior to 0.10.7 

[cli_failure.txt](https://github.com/user-attachments/files/30655468/cli_failure.txt)

### Output

```shell

```

ian1182:
The actual reolink-cli and reolink-gateway modules are fine as I could manually copy to  the windows "C:\Users\username\ .local\bin" folder

reolink-net:
You are right that this is a regression, and it is ours. Fixed on `main` — please re-run the command below and it should work now.

```
powershell -NoProfile -Command "iwr https://raw.githubusercontent.com/reolink/reolink-cli/main/install.ps1 -UseBasicParsing | iex"
```

Nothing to download or upgrade first: `install.ps1` is served from `main`, so the fix is already live for you.

## What broke

Your log named it exactly:

```
iex : Unable to find type [System.IO.Compression.ZipFile].
```

0.10.7 added an archive-layout check to `install.ps1` — part of hardening the installer after a privately reported issue — and that check calls `[System.IO.Compression.ZipFile]`. **Windows PowerShell 5.1 does not load that assembly until something asks for it**, while PowerShell 7 has it already. So a change meant to make the installer safer instead made it abort, on the shell that ships with Windows.

Note where your log stops: the download and the checksum verification both succeeded. It failed immediately after, on the first line of the new code. `Add-Type -AssemblyName System.IO.Compression.FileSystem` now runs first.

We had flagged `install.ps1` as parse-checked but not executed on Windows when that change shipped. Parsing was never going to catch a type that resolves at runtime, and this is what that gap cost. It should not have shipped that way.

## A second one, found while reproducing yours

On the Windows 11 machine used to reproduce this, the documented command failed *before* your error, at the download itself — hanging, or reporting `Access is denied`. `Invoke-WebRequest` on Windows PowerShell 5.1 hands the response to the Internet Explorer engine for DOM parsing unless told not to, and where IE's first-launch configuration never ran, it stalls:

| command | result |
|---|---|
| `iwr <url>` | hangs, no output |
[...]

PR #62, opened 2026-08-03
fix(windows): iwr needs -UseBasicParsing, or the install stalls before it starts (#60)

cuican-reolink:
Reported as #60 and reproduced on a Windows 11 / PowerShell 5.1 machine.

`Invoke-WebRequest` on Windows PowerShell 5.1 hands the response to the Internet Explorer engine for DOM parsing unless told otherwise. Where IE's first-launch configuration never ran — most machines now — that stalls indefinitely or fails with `Access is denied`, **before anything is downloaded**. The error names neither Internet Explorer nor the flag that fixes it.

Measured on the failing box:

| command | result |
|---|---|
| `iwr <url>` | hangs, no output, killed at 600s |
| `iwr <url> -UseBasicParsing` | `OK len=9091` |

On PowerShell 7 the parameter is already the default, so the same line works everywhere.

Verified by running the exact command now in the README, from cmd.exe, on the machine that could not install before — download, checksum verified against the pinned repository, both binaries installed, `reolink-cli 0.10.7 (external · LAN-only)`.

Second Windows defect in a row that a parse check could not catch; both only appear at runtime on 5.1.


reolink-net (approved):
Reproduced and fix verified on the reporting platform before approving. Merging promptly — main serves this script live and Windows users are currently blocked.

PR #63, opened 2026-08-03
docs: the one-line install had no documented way to uninstall

cuican-reolink:
Found while running the Windows paths that had never actually been executed.

The uninstall section assumed a release tarball and its `uninstall.sh`/`uninstall.ps1`. Anyone who used the one-line installer has **no tarball**, so they had no documented way to uninstall — while the binary has carried `setup --uninstall` all along. That is now the primary instruction.

It also documents the Windows behaviour instead of leaving it to be discovered: the uninstall reports that it could not remove `reolink-cli.exe` and exits non-zero, because Windows locks a running image and the binary doing the uninstalling cannot delete itself. The message carries the command that finishes the job.

Exiting non-zero there is deliberate and new — it previously exited **0** having left the main binary behind, which told every script, package manager and agent reading the exit code that the uninstall had succeeded. That fix, along with a scoping bug where uninstalling from one prefix killed every `reolink-gateway.exe` on the machine, is in the binary and ships with the next release.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -148,7 +148,23 @@ Slash commands run a single deterministic shell command — **no LLM intent mapp
 
 ## Uninstall
 
-The release tarball ships its own `uninstall.sh` / `uninstall.ps1` scripts:
+**Installed with the one-line installer?** There is no tarball, so use the
+binary's own uninstall — this is the path most installations need:
+
+```sh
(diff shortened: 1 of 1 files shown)


reolink-net (approved):
The one-line install genuinely had no documented uninstall path; making the built-in command primary is right. Windows non-zero exit is documented as expected rather than left to surprise people.

PR #64, opened 2026-08-03
security: document the DNS-rebinding defence, and guard every past report mechanically

cuican-reolink:
Two things, both from worrying that a careful reporter would come back to the same ground.

## The cross-origin guard was bypassable by DNS rebinding

It compared `Origin` against `Host`. Rebinding makes those agree: the attacker serves `evil.example`, lets the TTL lapse so the browser re-resolves it to 127.0.0.1, and the page fetches `http://evil.example:9000/…`. Both headers are then `evil.example:9000` — matching — so the guard allowed it and `ACAO: *` handed the response back.

Not merely a read: `POST /api/cameras/<name>/login` authenticates with the password stored in `aliases.toml`, so any page could mint a token and then read snapshots and the live event stream — the exact threat the guard's own comment described.

The gateway now also requires `Host` to be an address it actually bound to. Rebinding depends on a *name*, and a name is never the address we bound. Both checks run only for requests carrying an `Origin` — curl, go2rtc and Home Assistant server-side pulls send none and are unaffected, hostname or not.

A probe test was written first to prove the old logic really did admit the request, then the fix. Deliberate consequence, documented: a browser reaching the dashboard through a custom hostname is refused, because server-side that is indistinguishable from the attack.

*(The binary change ships with the next release; this PR carries the policy and the docs.)*

## Every past report is now an assertion

`scripts/check-past-findings.py` — 33 assertions covering fifteen documentation findings and both security advisories. Each asserts the **property that was wrong**, never the wording of the fix: a check pinned to today's phrasing fails on the next edit, and a checker that cries wolf gets ignored, which is worse than not having one. Verified by fault injection — seven classes broken one at a time, every one caught, green again on restore.

Writing it found two live regressions:

| | |
|---|---|
| `install.ps1`'s own `Invoke-WebRequest`/`RestMethod` calls | still lacked `-UseBasicParsing` — #60 was fixed in the documented one-liner but not inside the script, leaving the same PowerShell 5.1 stall one code path away |
| `rm -rf ~/.cache/reolink` in SKILL.md and voice-alert.md | missing the `-cli` suffix, so it removes a directory that does not exist and leaves the cache untouched — same defect as #13, in two places nobody had checked |

Also records that gateway session tokens expire after 300 s of inactivity, which was true in the code and missing from the policy.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -91,6 +91,28 @@ Knowing the intended behaviour makes it easier to tell a bug from a feature:
 - The gateway listens on `127.0.0.1` by default and **refuses browser
   cross-origin requests**. Binding it to `0.0.0.0` is an explicit opt-in and
   exposes camera control to your whole LAN.
+
+  Loopback is not a boundary against a web page — your browser can reach
+  `127.0.0.1`, and the gateway logs in with stored credentials. Two checks
+  apply to any request carrying an `Origin`, which means browsers:
+
+  1. the `Origin` must match the `Host` it addressed, and
+  2. that `Host` must be an address the gateway actually bound to.
+
+  The second exists because DNS rebinding defeats the first: an attacker who
+  serves `evil.example`, lets its TTL lapse and has the browser re-resolve it to
+  `127.0.0.1` produces a request where `Origin` and `Host` are *both*
+  `evil.example` — perfectly matching, and refused anyway because the gateway
+  never bound that name.
+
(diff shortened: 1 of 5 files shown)


reolink-net (approved):
The rebinding bypass is real and was demonstrated with a probe test before the fix. The past-findings checker was fault-injected across seven classes rather than assumed to work, and it caught two live regressions on its first run — which is the argument for it.

PR #66, opened 2026-08-04
security: keep GITHUB_TOKEN off argv, and stop the checksum check passing on an unreadable file

cuican-reolink:
Two findings from an adversarial audit run through the reporting contributor's methodology. Both are the project disagreeing with itself.

**`GITHUB_TOKEN` was on curl's command line** at three call sites in `install.sh`. `self_update.rs` refuses to do exactly that and explains why in a comment: argv is world-readable through `ps`, so any other local user can lift the token for as long as the request runs. Same project, same threat, opposite answers — with the reasoning already written down on one side.

The header now goes to curl on stdin. It is fed by `printf | curl` rather than curl reading the script's own stdin, because this file is routinely run as `curl … | sh` where the shell's stdin **is** the script; a command reading from it would eat the rest of the install.

Verified against the live release:

| path | result |
|---|---|
| no token (normal public install) | installs 0.10.7, checksum verified |
| deliberately invalid token | GitHub returns 401 — which is itself the proof the header is being sent |

**`check-version-sync.sh` passed on an unreadable anchor.** For a checksums file whose contents no longer parse it printed `ok  committed hashes vs release  0 assets match what is served` and the run went green. Zero entries is not agreement — it is an unreadable anchor, and every one-line install using it fails closed because the installer finds no entry for its asset. A checker that reports success for the exact state it exists to catch is worse than not having one.

Now a MISMATCH, verified by injecting a corrupt file (run fails), then restoring (run green).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -77,8 +77,21 @@ if [ "$os" = linux ]; then
 fi
 
 # Optional bearer token (private repo only); public access is anonymous.
-AUTH=""
-[ -n "${GITHUB_TOKEN:-}" ] && AUTH="Authorization: Bearer $GITHUB_TOKEN"
+#
+# The header goes to curl on **stdin**, never on the command line. argv is
+# world-readable through `ps`, so any other user on the machine could lift
+# GITHUB_TOKEN for as long as the request runs. `self_update.rs` already refuses
+# to do this and says why; this script was doing the opposite at three call
+# sites, in the same project, against the same threat.
+#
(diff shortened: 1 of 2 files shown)


reolink-net (approved):
Both verified against the live release before approving. The token change was checked on the anonymous path (installs) and the token path (401 from GitHub, proving the header is sent); the checksum check was fault-injected and restored.

PR #67, opened 2026-08-04
release: 0.10.8

cuican-reolink:
Version strings, the CHANGELOG entry, and the six artifact checksums.

The checksums land on the default branch **before** the release is published: both installers and `self-update` fail closed when the committed checksum file is missing, so publishing first would break every one-line install until this merged.

A security release — most of it found by auditing our own code rather than by a report, and one item introduced by the previous release's hardening. See the CHANGELOG entry for the full list; the headline items are `self-update` verifying nothing at all, the gateway's cross-origin guard being bypassable by DNS rebinding, and the event stream never re-checking its token.


reolink-net (approved):
Version strings agree across all seven manifests, the CHANGELOG entry is dated and covers every unreleased commit, and the six checksums were generated from the artifacts that will be uploaded. Checksums land before the release, as the fail-closed installers require.

Issue #68, opened 2026-08-04
`config init`: generated config tells users to copy it to `reolink/` (legacy dir); the tool reads `reolink-cli/`, and no copy is needed


ch-bas:
### What

The config that `config init` generates tells the user to copy it into a directory the tool does not read, and to perform a copy that has already been done.

Running `reolink-cli config init` (v0.10.8) writes `config.toml` and `aliases.toml` into the **active** config directory — on macOS `~/Library/Application Support/reolink-cli/`. But the header baked into the generated `config.toml` says:

```
# Example config.toml
# Copy to your platform config directory as reolink/config.toml.
# Typical paths:
#   Linux:   ~/.config/reolink/config.toml
#   macOS:   ~/Library/Application Support/reolink/config.toml
#   Windows: %APPDATA%\reolink\config.toml
```

`aliases.toml` carries the same `reolink/` guidance.

Two things are wrong with that header:

1. **Wrong directory.** It names `reolink/`, but the tool's config directory is `reolink-cli/`. `doctor` confirms both facts — the active dir and that `reolink/` is the *legacy, migrated-away* location:

   ```
   [ OK ]  config directory       …/Library/Application Support/reolink-cli
   [ OK ]  legacy reolink/ dirs   none left behind (clean post-migration)
   ```

   A user who follows the header copies their config to `~/Library/Application Support/reolink/` (or `~/.config/reolink/`), where nothing reads it — then wonders why their settings have no effect.

2. **The copy isn't needed at all.** `config init` already wrote the files to the active location. The "Example … Copy to …" framing reads like a sample the user must relocate, so at best it's busywork and at worst (combined with #1) it moves a correctly-placed file to a dead directory.

### Repro (override-free, isolated HOME so it doesn't touch your real config)

```sh
h=$(mktemp -d)
HOME="$h" reolink-cli config init
# → writes  $h/Library/Application Support/reolink-cli/config.toml   (note: reolink-cli)

grep -n 'Copy to\|reolink/' "$h/Library/Application Support/reolink-cli/config.toml"
# → # Copy to your platform config directory as reolink/config.toml   (note: reolink, no -cli)

HOME="$h" reolink-cli doctor --output text | grep 'config directory'
# → config directory   $h/Library/Application Support/reolink-cli
```

### Suggested fix

The header text is generated by the binary, so this is a binary change (filing as an issue per CONTRIBUTING, not a PR). Either:

[...]

PR #71, opened 2026-08-04
fix: devices slash command pointed at the legacy ~/.config/reolink path

ch-bas:
## What changes

The `/reolink-cli:devices` slash command described the registry as living at `~/.config/reolink/aliases.toml`, but the tool's config directory is `~/.config/reolink-cli/` — `reolink/` is the legacy, migrated-away path (`doctor` labels it "legacy reolink/ dirs … clean post-migration"). Corrected the description to `~/.config/reolink-cli/aliases.toml`.

This is the repo-side half of #68. The binary-generated config template still names `reolink/` too, but that string is compiled into the binary and can only be fixed upstream (tracked in #68); this PR fixes the one instance that lives in this repository.

## How it was verified

Grepped the whole repo after the change: `commands/devices.md` was the last doc still using the `reolink/` path — every other reference (AGENTS.md, SKILL.md, uninstall.md, the reference docs) already uses `reolink-cli/`. Frontmatter-only change to a slash-command description; no command behavior affected.

## Checklist

### DIFF
--- a/commands/devices.md
+++ b/commands/devices.md
@@ -1,5 +1,5 @@
 ---
-description: List registered Reolink aliases (from ~/.config/reolink/aliases.toml)
+description: List registered Reolink aliases (from ~/.config/reolink-cli/aliases.toml)
 ---
 
 Run `reolink-cli device list --output json` and print each alias on one line as:


reolink-net (approved):
Correct, and it is the last one. `reolink/` was the pre-0.8 config directory; the tool migrates away from it on startup and no longer reads it, so the description was pointing readers at a path where nothing they wrote would take effect.

Verified the rest of the repository is already consistent — `commands/devices.md` was the only file still naming the old path.

The binary-side half you noted in #68 (the `config init` template, the event-monitor `--help`, and the instructions the MCP server hands agents at startup) is fixed upstream and ships in 0.11.0.

Thanks for splitting it out this way.

PR #75, opened 2026-08-05
fix(plugin): installing the plugin gave Claude Code and Gemini zero MCP tools

cuican-reolink:
The README puts **39 MCP tools** in its headline. `.claude-plugin/plugin.json` declared only `"skills"` — no `mcpServers` — and there was no `.mcp.json` anywhere. `mcp/mcp.json` existed, but a repository-wide grep finds nothing that reads it: a dead file shipping in every release archive.

So the install path the README pushes hardest —

```
/plugin marketplace add reolink/reolink-cli
/plugin install reolink-cli@reolink-cli
```

— gave users the skill and **none of the tools**. `gemini-extension.json` had the same shape, carrying only `contextFileName`, so Gemini users got a markdown file instead of typed tools.

Per the plugin reference, a plugin declares an MCP server either inline in `plugin.json` or in a `.mcp.json` at the plugin root. Inline here: one file rather than a new one, and it cannot be confused with the separate project-scoped `.mcp.json` mechanism.

**The dead file's `env` block was deliberately not carried over.** It passed `REOLINK_PASSWORD` — which the MCP server itself refuses to accept in tool calls, because it would land in an agent transcript. It named `REOLINK_ALIAS` and `REOLINK_ALIASES_FILE`, both renamed several releases ago. And what `\${VAR}` expands to when unset is undocumented. A subprocess inherits the user's environment anyway, so omitting the block is simpler and more correct.

Verified by running the server the way the manifest starts it:

```
initialize ok — reolink-cli
tools/list ok — 39 tools
  camera_ping, camera_login, camera_info, camera_capabilities,
  camera_preview_capture, camera_snapshot, …
```

Repackaging confirms the archive's manifest now carries `mcpServers` and no longer ships `mcp/`.


reolink-net (approved):
Confirmed the file was orphaned before merging: nothing in the repository reads mcp/mcp.json. Handshake verified against the manifest's own invocation — 39 tools listed. Dropping the env block is right; it carried REOLINK_PASSWORD and two renamed variables, and relied on undocumented substitution.

Issue #74, opened 2026-08-04
vod download --type filter unimplemented


djancak:
### Version

0.10.8 (external · LAN-only)

### Platform

Linux x86_64

### Camera model and firmware

RLC-520A v3.2.0.5180_2507241369 && RLC-810A v3.1.0.956_24022103

### Command you ran

```shell
❯ reolink-cli -c my-cam vod search --from "2026-08-04T00:30:00" --to "2026-08-04T05:50:00" --type people | jq | grep recordType

and also...

❯ reolink-cli -c my-cam vod search --from "2026-08-04T00:30:00" --to "2026-08-04T05:50:00" --type md,people | jq | grep recordType

and also...

❯ reolink-cli -c my-cam vod search --from "2026-08-04T00:30:00" --to "2026-08-04T05:50:00" --type "md,people" | jq | grep recordType
```

### What you expected, and what happened instead

I expected to receive only results with `"recordType": "people"` or  `"recordType": "md,people"`.
However, all `"recordType": "md"` results were received, including `"recordType": "md,people"`

### Output

```shell
"recordType": "md",
"recordType": "md",
"recordType": "md",
"recordType": "md",
"recordType": "md,people",
"recordType": "md",
"recordType": "md",
"recordType": "md",
```

ch-bas:
Confirming the characterization from the docs/CLI side (I don't have hardware to repro the device behaviour, but this narrows where the fix lives).

The flag is advertised as a filter, not just accepted. `vod search --help` on 0.10.8:

```
--type <RECORD_TYPES>  Recording type filter (comma-separated: people,vehicle,md,...)
                       [possible values: manual, sched, io, md, people, vehicle, face, dog_cat, visitor, other, package]
```

So the value is a validated enum (a typo'd type is rejected at parse time), yet your output shows pure `"recordType": "md"` entries coming back under `--type people` — i.e. the filter isn't being applied to the result set at all, regardless of value. That rules out a usage error: `people`, `md,people`, and `"md,people"` are all valid syntax for this flag.

[...]

Issue #73, opened 2026-08-04
Is it possible to add a armv7 build


ian1182:
### What are you trying to do?

Run the reolink-cli package on a Raspberry Pi ZeroW or indeed any other Pi running a 32-bit OS.  A one-time armv6 & armv7 build was provided via the discord engineers-table forum for 0.10.1 and that worked. Whilst the armv6 may be helpful for very old Pi models I suspect armv7 is more common.

### How do you do it today?

Use the older 0.10.1 build supplied via the discord channel but this lacks the updates in the subsequent releases.

### Camera models this applies to

N/A

reolink-net:
Done in [v0.11.0](https://github.com/reolink/reolink-cli/releases/tag/v0.11.0) — `linux-armv7` and `linux-armv6`.

Two archives, not one: a Pi 1 or Zero cannot execute armv7 code at all. `install.sh` and `self-update` each resolve the archive matching the build they are replacing rather than guessing from the CPU, so a machine keeps getting its own kind.

Built against glibc 2.28, so Raspberry Pi OS Buster and later work — the default toolchain target would have been 2.34 and excluded everything before Bookworm.

The reason these were not published before was that nobody here had the hardware. That is now covered by emulation: armv6 verified on Raspberry Pi OS armv6 (`uname -m` = `armv6l`), armv7 on Debian Buster and Bullseye, both through the real one-line install, plus a negative control confirming the armv7 binary refuses to run on an ARMv6 CPU.

One caveat worth knowing: 32-bit Raspberry Pi OS has booted a 64-bit kernel by default since Bullseye, so `uname -m` says `aarch64` while the userland is 32-bit. The installer reads `getconf LONG_BIT` instead and picks armv7 for that case.

Issue #70, opened 2026-08-04
Design discussion: `config init` seeds enabled example cameras that join `--all-devices` / tag fan-outs


ch-bas:
> **Design discussion, not a bug report.** (GitHub Discussions isn't enabled on the repo, so I'm raising it here.) Everything below works as documented — this is a judgment call about a default, and I'd like your view before anything is changed.

### The tension

`config init` seeds `aliases.toml` with example cameras so the file format is self-documenting — which is genuinely helpful. Two of them ship **enabled** and carry routable-looking private addresses and shared tags:

```toml
[cameras.front-door]           # enabled
host = "192.168.1.41:9000"
tags = ["outdoor", "entry"]

[cameras.garage]               # enabled
uid = "952700XXXXXXXXXX"
tags = ["outdoor", "garage"]

[cameras.lab-v30]              # disabled = true
host = "192.168.1.88:9000"
tags = ["lab", "experimental"]
```

Because `front-door` is enabled and tagged `outdoor`, a brand-new user who runs a fan-out **before** pruning the examples —

```sh
reolink-cli --all-devices ping
reolink-cli --tag outdoor snapshot
```

— sends a real TCP connection attempt to `192.168.1.41:9000`, an address they never chose and which may host an unrelated device on their LAN. The README does say to remove the placeholders, but they look real in `device list` (the only tell is `hasPassword:false`), and fan-out selectors are exactly the "just try everything" commands a new user reaches for first.

This sits a little awkwardly next to the project's own **Responsible use** guidance ("Not a scanner… do not use `discover`/`ping` to sweep networks you were not asked to work on"): the default config primes precisely that kind of unsolicited probe.

### Why I think it's worth a look rather than nothing

You already treat "disabled" as the safety valve — **`lab-v30` ships `disabled = true`** so it won't participate in fan-outs. `front-door` and `garage` don't, so the safety is applied inconsistently to your own seed data.

### Options (not advocating strongly for any)

1. **Ship all example entries `disabled = true`.** They still document the format and still show in `device list`, but `--all-devices` / `--tag` skip them until the user flips `disabled` (or replaces them). Smallest change, internally consistent with `lab-v30`.
2. **Use unambiguously non-routable example addresses** — RFC 5737 `192.0.2.x` (TEST-NET-1) instead of `192.168.1.x` — so a stray fan-out can't hit a real neighbour.
[...]

Issue #69, opened 2026-08-04
`--output text` renders JSON (not a table) for `device list`, `device show`, `cache status`


ch-bas:
### What

`--output text` is honored by some commands and silently ignored by others, which is exactly the commands a human is most likely to want as a table.

`doctor`, `status`, and `features` render clean human-readable output under `--output text`. But the registry / data commands emit a status line followed by pretty-printed **JSON** — not text:

```
$ reolink-cli device list --output text
device: ok
[
  {
    "camera": "front-door",
    "channel": 0,
    "description": "Front entrance",
    ...
```

Measured on v0.10.8 (offline commands only):

| command | `--output text` produces |
|---|---|
| `doctor` | text table ✓ |
| `status` | text dashboard ✓ |
| `features` | text ✓ |
| `device list` | **`device: ok` + JSON** |
| `device show <name>` | **`device: ok` + JSON** |
| `cache status` | **`cache: ok` + JSON** |

So a user who discovers `--output text` from `doctor` and applies it to `device list` — the single most natural command to want as a table — still gets JSON, just with the envelope stripped and a `device: ok` line prepended.

### Repro

```sh
reolink-cli device list --output text     # status line + JSON, not a table
reolink-cli doctor --output text          # a real text table, for contrast
```

### Expected

Either render these as a table under `--output text` (a `device list` table of name / host-or-uid / tags / hasPassword would be the high-value one), or, if a text renderer isn't planned for a given command, at least don't prepend a `<cmd>: ok` line to raw JSON — that combination reads as "text mode is supported here" while delivering JSON.

The rendering lives in the binary, so filing as an issue per CONTRIBUTING rather than a PR. Low severity — a UX consistency gap, nothing is broken. Environment: reolink-cli 0.10.8 (external · LAN-only), macOS arm64.

reolink-net:
Fixed in [v0.11.0](https://github.com/reolink/reolink-cli/releases/tag/v0.11.0), and fixed in the renderer rather than per command — the next list command gets a table without anyone remembering to add one.

```
$ reolink-cli device list --output text
device: ok
camera      channel  description      disabled  hasPassword  host               protocol  tags
----------  -------  ---------------  --------  -----------  -----------------  --------  -------------
front-door  0        Front entrance   true      false        192.168.1.41:9000  v20       outdoor,entry
```

[...]

PR #77, opened 2026-08-06
release: 0.11.0

cuican-reolink:
Sync for the 0.11.0 release. **Merge this before publishing the release** — the installers resolve the latest release and then look for its checksum file here, so a release published first fails closed for everyone until this lands.

## What is in it

| | |
|---|---|
| Version | 7 manifests to 0.11.0 |
| Checksums | `checksums/v0.11.0.sha256`, 8 assets, written on the build machine |
| Changelog | 0.11.0 entry |
| Docs | platform list, uninstall guidance in `AGENTS.md` / `SKILL.md` / `commands/uninstall.md` |
| Installer | `install.sh` learns `armv7l` / `armv6l` |
| Tooling | `scripts/check-doc-commands.py` no longer executes what it checks |

## Reported issues fixed

- **#74** — `vod search --type` did not filter. The request carried the type all along and the camera ignored it, so `--type people` returned every motion recording in the window. Filtering now happens on our side, which fixes it for the MCP tools and the HTTP API too. A recording carries a set of tags and so does the request, so `--type people` keeps one tagged `md,people`.
- **#69** — `--output text` printed JSON for `device list`, `device show` and `cache status`. Fixed in the renderer rather than per command, so future list commands are covered by the same change.
- **#68** — `config init` told users to copy its output into `reolink/`, which the tool does not read, and the copy had already been done. The same stale path was in the event-monitor `--help` and in the instructions the MCP server hands every agent at startup.
- **#70** — raised as a design question about the seeded examples; the answer turned out to be a bug. `disabled = true` was honoured only by `device list`, so `--all-devices` and `--tag` still contacted disabled cameras. A fresh install running `--all-devices ping` opened TCP connections to `192.168.1.41`. Fan-outs now skip disabled cameras; naming one with `--camera` still works, because asking for a camera by name is intent.
- **#73** — 32-bit Raspberry Pi builds. Two archives, not one: a Pi 1 or Zero cannot execute armv7 code, and `self-update` resolves the archive matching the build it is replacing rather than guessing from the CPU.

**#71 is not included here.** It fixes `commands/devices.md`, the repo-side half of #68. It is the same one-line change, authored by @ch-bas, and it should be merged on its own so the credit stays where it belongs. Either order works; there is no conflict.

## Behaviour change: `setup --uninstall`

The command had no confirmation of any kind — not a prompt, not a flag check. With `--purge` it removes both binaries, the config, the camera registry, the agent skill directories and the Claude Code plugin registration, and it needs no config, no credentials and no terminal to do it.

It now refuses when stdin is not a terminal unless `--no-interactive` is present. That flag has been in `AGENTS.md` all along and nothing read it, so the documentation described a confirmation step that did not exist.

**If you script an uninstall, add `--no-interactive`.** The bundled `uninstall.sh` / `uninstall.ps1` wrappers and the slash commands pass it through. A person at a terminal is unaffected, including when stdout is redirected — the check reads stdin.

## Verification

- 643 tests, 0 failures; clippy clean
- Every documented command parses (296 of them), checked with a script that can no longer execute what it checks
- All 33 assertions from previous reports still hold
- 8 artifacts, all past the P2P fingerprint gate, with an internal build as the positive control (81 hits) so a zero means something
- armv7 on Debian Buster and Bullseye; armv6 on Raspberry Pi OS armv6 (`uname -m` = `armv6l`), including the negative control that an armv7 binary refuses to run there
- Windows: the uninstall gate refuses, and `--parse-only` validates without executing
- `--type` filtering checked against a live camera, in both directions


reolink-net (approved):
Release sync for 0.11.0. The checksums are written on the build machine and cover all eight archives; they have to land before the release is published, or the one-line installers fail closed in the interval.

Issue #72, opened 2026-08-04
Allow multiple vod downloads in one command, or keep connections alive between commands.


djancak:
### What are you trying to do?

right now, only one file can be downloaded per execution of the program
such as with `reolink-cli -c my-cam vod download 0120260803025316 --file /tmp/whatever.mp4`
when fetching many video files using this in a script, the delay time between downloads is very noticeable. 

I presume that this delay happens because the program must re-establish a connection to the camera every time that it is executed to perform an action with the camera.

If all the files could be downloaded with one command and one connection, that would be spiffy.
something like `reolink-cli -c my-cam vod download 0120260803065316,0120260803065332,0120260803065316,0120260803065335 --file /tmp/whatever.mp4,"/tmp/unfortunate file,with,commas and spaces.mp4",/tmp/whatever2.mp4,/tmp/whatever3.mp4` could work, though it looks ugly.

Could also just keep the connection alive inside the `reolink-gateway`, assuming I understand the purpose of a gateway. Each successive `reolink-cli vod download` should pipe quickly through the gateway to get data flowing steady across the network. Maybe a `--keep-alive` flag? Would be useful for development of a TUI, too! Imagine a Reolink TUI using the Kitty Graphics Protocol over ssh!

### How do you do it today?

I use a Python script with a reverse engineered Baichuan protocol class alongside the official `reolink_aio` module developed by StarKillerOG. It downloads hundreds of videos over a single connection with hardly a dip in network traffic between files, saving minutes of time!

### Camera models this applies to

i used a RLC-810A and RLC-520A, but seems like it would apply to all models.

reolink-net:
Done in [v0.11.0](https://github.com/reolink/reolink-cli/releases/tag/v0.11.0) — your first suggestion, several names in one command:

```bash
reolink-cli -c my-cam vod download 0120260803025316 0120260803025332 0120260803025401
```

Your diagnosis was right. A download deliberately avoids the shared session — a transfer running for minutes would stall every other command against that camera — so each invocation paid its own connect and login. Measured at ~54 ms device-side, about a quarter of a second end to end per file.

I did not add `--keep-alive`. Holding the connection between invocations needs idle timers and eviction rules to answer a question batching already answers, and a flag whose effect you cannot see is a poor interface. If you have a case batching does not cover, reopen — the reasoning above is the whole of it, not a policy.

[...]

PR #78, opened 2026-08-06
docs(changelog): 0.11.0 fixes six reported issues, not five

cuican-reolink:
The count in the 0.11.0 heading was written before #72 was folded into the release. The entry below it lists #68, #69, #70, #72, #73 and #74 — six.


reolink-net (approved):
Correct — the entry lists six.

Issue #79, opened 2026-08-11
`credentials.key` and `doctor` ignore `--cameras-file` / `--config-file` — the key lands in the default config dir, and doctor diagnoses files the CLI is not using


ch-bas:
## Summary

The global `--cameras-file` / `--config-file` flags (and `REOLINK_CAMERAS_FILE` / `REOLINK_CONFIG_FILE`) are honored by every data-file read and write I tested — but two things stay anchored to the platform-default config directory regardless:

1. **`credentials.key`** is always read/written in the default config dir, never beside the redirected registry. README and SKILL.md both say the key lives "beside `aliases.toml`" and that backing up means "copying **both** files" — for a redirected setup that advice copies the wrong pair, and the restore ends in *"cannot be decrypted — the key file is missing"*. Reproduced end-to-end below.
2. **`doctor`** accepts both flags but its `config.toml`, `cameras registry`, and `registry perms` checks always inspect the default paths — while the same command *does* honor `--gateway-addr` for the gateway probe. Under redirection it reports "cameras registry … not found → run `device add`" while `device list` happily reads three cameras from the redirected file, and the perms check no-ops ("registry not yet created") instead of checking the mode of the registry actually in use.

Both look like the same root cause: one path resolver for data files, another for the key file and doctor.

## Reproduction (0.11.0, darwin-arm64, checksum-verified release binary)

Key location — sandboxed `$HOME`, everything redirected:

```bash
H=$(mktemp -d); mkdir -p "$H/custom"
HOME=$H REOLINK_CAMERAS_FILE=$H/custom/aliases.toml REOLINK_CONFIG_FILE=$H/custom/config.toml \
  sh -c 'printf "<camera-password>\n" | reolink-cli device add cam-x --host 192.0.2.20:9000 --user admin --password-stdin'

ls "$H/custom"
# aliases.toml            ← ciphertext (RLENC1:…), as expected
ls "$H/Library/Application Support/reolink-cli/"
# credentials.key         ← the key went to the DEFAULT dir, not beside the registry
```

Backup per the docs, restored on a "new machine" (fresh `$HOME`):

```bash
B=$(mktemp -d); mkdir -p "$B/custom"
cp "$H/custom/aliases.toml" "$B/custom/"     # "copying both files" — but there is no key next to the registry
HOME=$B REOLINK_CAMERAS_FILE=$B/custom/aliases.toml reolink-cli --camera cam-x login
# warning: the stored password for cam-x could not be decrypted — the key file
[...]

PR #81, opened 2026-08-14
release: 0.12.4

cuican-reolink:
0.12.4: a reported data-loss defect in redirected setups, a protocol default that stops guessing, and the Wi-Fi switch work that came out of a router compatibility run.

**Fixed for a reporter:** #79 — `credentials.key` was always created in the platform default config directory, so `--cameras-file` / `--config-file` produced a registry whose key never travelled with it. The documented backup ("copy both files") was impossible to follow in that layout and restoring on another machine lost every stored password. `doctor` had the same split: it forwarded `--gateway-addr` but not the other two, so it diagnosed files the CLI was not using. The key now lives beside the file whose passwords it protects — which makes the documented advice true in every layout and gives each redirected profile its own key — and `doctor` inspects the resolved paths. Installations that predate the change keep working: the old location is still read, read-only.

**Contract change:** the wire protocol now defaults to `v20` instead of probing for it. The probe cost an extra TCP round trip that, at its old 750 ms budget, was shorter than the OS's first SYN retransmission — a single lost SYN on Wi-Fi turned a reachable camera into `route not found: TCP connect timed out`. v30 cameras declare `protocol = "v30"`; `--protocol auto` still forces detection. `wifi set` also gained `phases:` / `switch:` output lines, and `data.new_host` now strictly means "the host written into the entry" (`null` when nothing was written) rather than "an address rediscovery happened to see".

**Wi-Fi:** `wifi set` measures the switch from client-side reachability instead of reading the SSID back — the v20 config command returns stored configuration, not association state, so the old number was a round trip rather than a transition. The "is this camera on Wi-Fi" gate now uses measured signal strength instead of `GetLinkType`, whose values (`LAN`/`PPPOE`/`CDMA`) describe how an IP is obtained and contain no wireless value at all, so wired cameras were being told to pass `--no-test`. `wifi get` reports `signal`.

`checksums/v0.12.4.sha256` covers all eight archives and is written from the build machine's `dist/SHA256SUMS`, not re-downloaded from the release. It has to land on `main` before the release is published, or the one-line installers fail closed for everyone in between.

All eight archives are external (LAN-only) builds; the P2P fingerprint gate reported `clean (external)` for each.


reolink-net (approved):
Release PR. Eight external (LAN-only) archives, P2P fingerprint gate clean on every one; checksums written from the build machine and cross-checked against the archives. Artifacts run-verified on macOS arm64 (issue #79 reproduction end to end, plus a live camera), Windows x64 (live camera, discover reporting remoteUid: 0), Linux x64/arm64 glibc and musl; armv6/armv7 build-verified only, as before. Merging this before publishing the release so the installers never resolve a tag whose checksum file is missing.

PR #82, opened 2026-08-14
docs: state the credentials.key pair rule explicitly for redirected setups

ch-bas:
## What changes

The docs half of #79, as offered there and taken up by the maintainer. The 0.12.4 binary makes `credentials.key` live beside the file whose passwords it protects; this PR makes README §"Safe credential handling" and the SKILL.md credential bullet *state* that rule instead of merely no longer contradicting it:

- the per-file rule, including redirected `--cameras-file` / `--config-file` setups (the layout where the old wording bit);
- back up the file **and the `credentials.key` next to it** as a pair, from the same directory — with a worked example of a redirected profile, since that is the exact step that used to lose every stored password;
- **separate registries are separate secrets** — per-profile key isolation, stated explicitly per the maintainer's comment in #79 ("we will say so explicitly in the docs rather than leave it implicit");
- the pre-0.12.4 fallback: a key still in the default config directory is read until the next write moves it beside its file;
- the loud failure mode on restore-without-key, so nobody mistakes it for a wrong password.

No other surface mentions the key — `GEMINI.md`, `copilot-instructions.md`, and `AGENTS.md` were grepped for `credentials.key` / encryption / backup wording and carry none, so this diff is the complete set.

## How it was verified

The failure mode the text documents was executed end to end in #79 against the checksum-verified v0.11.0 release binary (sandboxed `$HOME`s, redirected registry, restore without the neighbouring key → "cannot be decrypted — the key file is missing", exit 1, no network attempt).

The claims that describe **0.12.4 behavior** (key created beside a redirected registry, pair-copy restore decrypting, default-directory fallback read, doctor honoring the flags, two profiles → two distinct keys) follow the maintainer's confirmed design and the 0.12.4 CHANGELOG entry. The v0.12.4 release assets are not published at the time of opening; I will run every example in this diff against the checksum-verified 0.12.4 binary as soon as they are, and post the transcript as a comment here before this merges.

## Checklist

- [x] No password, camera UID, or private network address appears in the diff (placeholder paths only: `/srv/cams/site-a`, `/backup/site-a`)
- [ ] `cargo test --workspace` passes — N/A, no Rust in this repo


reolink-net:
v0.12.4 is published — assets and `checksums/v0.12.4.sha256` are both up, so the verification you committed to in the description can run whenever suits you. We are holding the merge for it rather than landing this first; the transcript is the part that makes the examples worth trusting, and it is yours to post.

https://github.com/reolink/reolink-cli/releases/tag/v0.12.4

Reviewed the text against the implementation, claim by claim, and it is accurate on all of them — including the two that are easy to get subtly wrong:

- **"a redirected `config.toml` (it can hold a password too)"** — correct, and it is the reason the rule is per-file rather than derived from the registry directory. A registry-only rule would have left exactly that case broken.
- **"a key still living there is read as a fallback until the next write moves it beside its file"** — correct, and the read-only half matters: the legacy location is never written back, so the first `encrypt` after upgrading puts the key where it belongs without stranding anything.

"Separate registries are separate secrets" is the sentence we owed you from #79. There is a unit test named after it.

One heads-up on `skills/reolink-cli/SKILL.md`, since it affects you specifically: that file is mirrored by hand into the internal repository the releases are built from, and on 2026-07-31 a release sync silently reverted a contributor's merged pull request that way — the fix landed here, the sync copied the older internal copy over it, and neither review nor CI noticed. Your wording is going back into the internal copy as part of merging this, not afterwards as a follow-up, so the next sync carries it rather than overwrites it. Flagging it because "merged" and "survives the next release" have not always been the same thing in this repo, and you should not have to know that.

No conflict with `main`: the protocol bullet directly below yours changed in 0.12.4 (the default is now `v20` instead of the magic probe) and your diff already sits on top of it.

reolink-net (approved):
Accurate on every claim, checked against the implementation and then executed against the published v0.12.4 binary rather than reasoned about. The one statement that needed runtime proof rather than a code read is "Separate registries are separate secrets": two profiles do get distinct keys, and crossing them (site-b's registry with site-a's key) fails with "could not be decrypted" while site-b's own pair resolves normally. That is a security-shaped sentence in customer docs, so it should not ship on a unit test alone.

reolink-net:
Merged — ahead of the transcript you offered, so this is us changing the plan we stated two days ago, not you missing a deadline.

The reason is that the text was already gating on us rather than on you: the docs on `main` described a rule the shipped binary no longer followed, and every day that sat there was a day a redirected user could read the old advice and lose passwords by following it. Waiting for a second confirmation of claims we could confirm ourselves was buying very little and costing that.

So we ran the verification in your place rather than skipping it. Against the published v0.12.4 macOS binary, in throwaway `$HOME`s, the way you would have:

- key created beside a redirected registry, default config directory never touched;
- pair-copy (registry + neighbouring key) restores on a fresh `$HOME` and decrypts;
- registry alone, no key → `cannot be decrypted — the key file is missing`, and the path named in the message is the one beside the redirected registry, not the default;
- a key left in the old default directory is still read, and is not written back;
- `doctor --cameras-file --config-file` reports the redirected files, including `registry perms` on the registry actually in use.

The one that needed running rather than reading was **"Separate registries are separate secrets"** — a security-shaped sentence in customer docs should not ship on a code read. Two profiles do get distinct keys, and crossing them fails closed:

```
site-b registry + site-a key  → "could not be decrypted"   (INPUT_ERROR)
site-b registry + site-b key  → ok, hasPassword=true
```

Your wording went into the internal copy of `SKILL.md` in the same operation as the merge, and `check-docs-mirror.sh` now reports the two repositories agree. That was the specific commitment from the previous comment, and it is the thing that failed you on 2026-07-31.

If your own run turns up anything that reads wrong, please do still post it — a follow-up PR on top is welcome and will not be treated as re-litigating a merged decision. #79 is closed by this and v0.12.4.

Thanks for the report and for the docs half. The reason this one moved quickly is that your issue arrived with a reproduction that made the diagnosis mechanical.

PR #84, opened 2026-08-21
docs(readme): add a Commands reference for supported features

cuican-reolink:
The README had a Highlights list and a lot of responsible-use / security disclaimers, but no concrete, browsable list of **what the CLI actually supports** — and the Highlights were already missing the new `encode` and `siren` capabilities.

This adds a **Commands** reference section (grouped table of every top-level command and what it does), placed right after Highlights. To keep it from drifting, the section leads with the canonical, always-current sources — `reolink-cli --help` and `reolink-cli features` (the machine-readable manifest agents read) — and frames the table as a browsing snapshot.

Also:
- Adds `encode` (video encoder settings) and `siren` to Highlights.
- Notes that a command existing does not mean a given model implements it (`capabilities` / `device inventory --capabilities` tell you per model), and that `v20` is the stable surface while `v30` is provisional.

Docs only — no code or release changes.


reolink-net (approved):
Docs: adds the Commands reference so supported features are discoverable.

Issue #87, opened 2026-08-22
Unable to download more than 210 videos per vod download request.


djancak:
### Version

reolink-cli 0.13.0 (external · LAN-only)

### Platform

Linux x86_64

### Camera model and firmware

RLC-810A

### Command you ran

```shell
[...]

Issue #86, opened 2026-08-22
remove arbitrary --limit of 500


djancak:
### What are you trying to do?

There are instances where I can have over 500 results from a vod search. 

`❯ reolink-cli -c cam-i vod search --from 2026-08-21T00:00:00 --to 2026-08-21T23:59:59 --limit 501
error: invalid value '501' for '--limit <LIMIT>': 501 is not in 1..=500`

It would be nice to be able to set `--limit 0` to get the full range of results from a query, because as of now I cannot rely on my scripts to process all camera files for a full day without doing a multi-query workaround.

### How do you do it today?

I just set `--limit 500` and cross my fingers that the results don't go over 500, which they did they other day.

### Camera models this applies to

all

reolink-net:
Done in [v0.13.1](https://github.com/reolink/reolink-cli/releases/tag/v0.13.1). `--limit` accepts up to 100000, and `--limit 0` means no limit.

You were right that it was arbitrary: the request carries no count at all and the camera pages until it runs out, so 500 was purely ours. Worse, it was indistinguishable from a real answer — `truncated: true` was the only hint. A day with 604 recordings and a month with 965 both come back whole now.

Issue #85, opened 2026-08-22
uid tag in aliases.toml breaks reolink-cli vod search (presumably other stuff, too!)


djancak:
### Version

reolink-cli 0.13.0 (external · LAN-only)

### Platform

Linux x86_64

### Camera model and firmware

RLC-810A v3.1.0.0 and RLC-520A v3.2.0.0

### Command you ran

```shell
❯ reolink-cli -c cam-i vod search --from 2026-08-21T00:00:00 --to 2026-08-21T23:59:59


❯ reolink-cli -c cam-ii vod search --from 2026-08-21T00:00:00 --to 2026-08-21T23:59:59
```

### What you expected, and what happened instead

Expected the vod search to work as in the previous version, but upon updating to 0.13.0 it fails.

I was able to resolve this issue by removing the `uid` tag in the `aliases.toml` file for `cam-i`, which is for the RLC-810A. I did verify that the `uid` was set correctly first by checking the web interface.

The RLC-520A that I had set up as `cam-ii` in the `aliases.toml` file did not have a `uid` set, so it continued to work normally.

### Output

```shell
Result for `cam-i` which had a `uid` set in the `aliases.toml` file:
{"ok":false,"command":"vod","protocol":null,"data":null,"error":{"code":"NETWORK_ERROR","message":"device unreachable: uid [ R E D A C T E D ]: wake timed out after 3 attempts: no response from device with uid [ R E D A C T E D ]","retryable":true}}

(it took about 9 seconds for it to fail)


Result for `cam-ii` which did NOT have a `uid` set in the `aliases.toml` file:
{"ok":true,"command":"vod","protocol":"v20","data":{"items":[],"scanned":0,"total":0,"truncated":false},"error":null}

(it is expected to get 0 results in this instance)
```

reolink-net:
Fixed in [v0.13.1](https://github.com/reolink/reolink-cli/releases/tag/v0.13.1) — you can put the `uid` back.

These builds carry no P2P, so a `uid` target went straight to a wake broadcast and ignored the `host` sitting in the same entry. A mains-powered camera never answers a wake, hence the ~9s timeout you measured. It now tries the LAN first, so an entry with both `uid` and `host` connects straight to the host; the wake path is left to the battery devices it was meant for. Verified against a real camera: 9.0s failure before, 2.1s success after.

Regression from 0.12.5, and it only showed up in the public builds — thanks for catching it.

PR #89, opened 2026-08-24
release: 0.13.1

cuican-reolink:
Release sync for **0.13.1**.

Lands `checksums/v0.13.1.sha256` **before** the release is published — the

### What is in 0.13.1

Three issues reported against 0.13.0 by @djancak:

- **#87** — `vod download` silently dropped everything past about the 210th
  recording. Recording names travel in the request URL and the gateway cut that
  URL off at 4096 bytes without saying so, then served the shortened request as
  a success. It now answers `414 URI Too Long`, and the CLI splits a long list
  into as many requests as it takes.
- **#85** — a camera registered with a `uid` could not be reached at all in
  these builds. A `uid` target went straight to a wake broadcast and ignored the
  LAN address the same entry already carried. It now tries the LAN first.
- **#86** — `vod search --limit` accepts up to 100000 and `--limit 0` means no
  limit. The old ceiling of 500 was never a camera limit, and one busy day can
  exceed it.

Two more defects surfaced while verifying #87 against a real Home Hub 2:

- `vod download` could not fetch any recording past the 500th of its day —
  locating a recording by name re-searched its day with a hard cap of 500.
- Long multi-recording downloads failed near the end with `invalid or expired
  token`; a media token expires 300s after its last use and a few hundred
  recordings stream for longer than that. Each batch now mints a fresh one.

And a performance change: locating a recording searched its whole day, but a
recording's name already ends in its own start time, so the search now asks for
that one second and falls back to the day only when that finds nothing. Same
220 recordings on the same camera: 360s → 204s.

### Verification

  `clean (external)` for every one.
- `reolink-cli --version` → `0.13.1 (external · LAN-only)`.
  per-artifact sidecars, 8/8.
- Fixes verified against a real Reolink Home Hub 2, not only unit tests:
  220-recording download completes 220/220, `--limit 0` returns 965 recordings
  where 500 used to truncate, and the `uid` failure reproduces at 9.0s before
  the fix and succeeds at 2.1s after.

The installer end-to-end runs (macOS + Alpine on both architectures) come after
the release is published, since they resolve the latest release.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
     {
       "name": "reolink-cli",
       "description": "Operate Reolink cameras locally via reolink-cli: discover, login, inspect, configure, preview, snapshot, and control PTZ/light/audio/detection/recording.",
-      "version": "0.13.0",
+      "version": "0.13.1",
       "source": "./",
       "author": {
         "name": "Reolink"
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "reolink-cli",
   "description": "Operate Reolink cameras locally: discover, login, inspect, configure, preview, snapshot, and control PTZ/light/audio/detection/recording.",
-  "version": "0.13.0",
+  "version": "0.13.1",
   "author": {
     "name": "reolink"
   },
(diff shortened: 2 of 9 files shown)


reolink-net (approved):
Checksum file lands before the release, as the installers require. All eight external artifacts built with the P2P gate reporting clean; fixes verified against a real Home Hub 2.

PR #90, opened 2026-08-26
release: 0.14.0

cuican-reolink:
Release sync for **0.14.0**.

Lands `checksums/v0.14.0.sha256` **before** the release is published — the

### What is in 0.14.0

**Scene mode** — the Home / Away / Disarm control a hub shows in the app, now on
the command line. A scene is a named set of per-channel tasks (`record`, `ftp`,
`email`, `push`, `audio` = the siren, `linkage`, `speaker`, `track`), so
switching scenes re-arms every paired camera at once. It does not change what
each camera *detects*; it changes what the hub *does* about it.

- `scene show` / `list` / `status` — which scene is active, what every scene
  holds, the running scene in full
- `scene set <id>` / `--schedule` — switch, or hand control to the timetable
- `scene edit` — tasks, name, icon, activation delay; `--channel N` confines the
  change to one camera
- `scene schedule get/set` — the weekly timetable (7 days × 2 half-hour slots ×
  24 hours)
- `scene alarm get/set` — which alarm types feed each task
- `scene options` — master switch, Home button, privacy mode while disarmed
- MCP tools: `camera_scene_show`, `camera_scene_list`,
  `camera_scene_schedule_get`, `camera_scene_set`

Scene id `0` is not a scene — it hands control to the timetable, reported as
`followSchedule: true`. Commands refuse rather than guess: an id the hub does
not have, a mistyped task name, or an option the device reports as unsupported
are all rejected with what the device actually offers.

Hubs and NVRs only — a standalone camera answers `405` to every `scene` command.

### Verification

  `clean (external)` for every one. Scanned the finished archives too: 0
  fingerprints and 0 bundled P2P libraries, against 170 fingerprints in an
  internal build used as a positive control — so the scan is known to work.
- `reolink-cli --version` → `0.14.0 (external · LAN-only)`.
  per-artifact sidecars, 8/8.
- **Verified against a real Reolink Home Hub 2 (v3.3.0.579)**, not only unit
  tests: every command and every flag was exercised on the device and the
  original configuration restored afterwards — scene switching (3 → 2 →
  timetable → 3), single-channel task edits, metadata edits, the weekly
  timetable, alarm-type mapping, and the four MCP tools.
- 737 tests pass. The XML fixtures are real device responses rather than
  hand-written samples, and two "silently does more than asked" invariants are
  pinned by their own tests: `--channel` touches only the named channel, and
  `--to` is exclusive.

The installer end-to-end runs (macOS + Alpine on both architectures) come after
the release is published, since they resolve the latest release.


reolink-net (approved):
Checksum lands before the release, as the installers require. All eight external artifacts built with the P2P gate clean and the finished archives scanned against a positive control; scene mode verified end to end on a real Home Hub 2.

Issue #94, opened 2026-08-27
Downloaded VOD extensions labelled .h264 when they're actually h265 codec


djancak:
### Version

reolink-cli 0.14.0 (external · LAN-only)

### Platform

Linux x86_64

### Camera model and firmware

RLC-810A and RLC-520A

### Command you ran

```shell
❯ reolink-cli -c cam-ii vod download 0120260826233104
{"ok":true,"command":"vod","protocol":"http","data":{"file":"0120260826233104.h264","bytes":12835157,"contentType":"application/octet-stream","codecFileName":"0120260826233104.h264"},"error":null}

❯ ffprobe -v error -select_streams v:0 -show_entries stream=codec_name 0120260826233104.h264
[STREAM]
codec_name=hevc
[/STREAM]
```

### What you expected, and what happened instead

Everything downloaded always has the `.h264` extension even if `h265`/`HEVC` is being used in the file. I suggest just switching to `.mp4` since using codec extensions implies a raw video stream without audio

reolink-net:
Confirmed, and the cause is worse than the extension — those files are not an
elementary stream at all. Reproduced on a Home Hub 2, first bytes off the wire:

```
00000018 66747970 69736f34 ...
         ^ftyp    ^iso4
```

That is an **MP4 container**, which is why `ffprobe` reads it fine and reports
hevc. So `.h264` was wrong twice over: wrong container, and wrong codec whenever
the recording was H.265.

The code assumed "frame 2 onward is the raw H.264/H.265 elementary stream" and
sniffed for Annex-B start codes. The MP4 header happens to contain a
`00 00 00 01`, so the sniffer read a garbage NAL type and fell through to its
`h264` default — which is why *everything* came out `.h264` regardless of codec.

Fixed for the next release: the container is detected first (`ftyp` at offset 4
→ `mp4`), and the Annex-B sniffing stays only as a fallback for firmware that
really does send an elementary stream. Downloads are now named `<name>.mp4`,
exactly as you suggested. Verified on device — same three recordings, before and
after:

```
0120260825194024.h264   magic=ftyp  ffprobe=hevc     ← before
0120260825194024.mp4    magic=ftyp  ffprobe=hevc     ← after
```

Thanks — the `ffprobe` output in the report is what made this quick to pin down.

reolink-net:
Released in [v0.14.1](https://github.com/reolink/reolink-cli/releases/tag/v0.14.1).

[...]

Issue #93, opened 2026-08-27
update --file for vod download to accept multiple inputs


djancak:
### What are you trying to do?

Presently (reolink-cli 0.14.0) `vod download --file` only works when downloading an individual file. It should be updated to accept as many arguments as `vod download`, and if the number of arguments between them do not match, then throw an error indicating the discrepancy. Perhaps also make a `--files` alias as well.

Alternatively, separate `--file` into two different options `--directory` and --`file(s)`/`filename(s)`
`--directory` would be the main directory under which all downloaded files are stored, just one argument.
`--files`/`--filename(s)` could still include sub-directories in the final output. use default names if this option is not set. would of course need to match the same number of arguments in `vod download`

so `reolink-cli -c cam-i vod download 0120260825224024 0120260825223942 0120260825221659 --directory /path/to/directory --files file1.mp4 "file 2.mp4" "/special record/file 3.mp4"` would output to:

```
/path/to/directory/file1.mp4
/path/to/directory/file\ 2.mp4
/path/to/directory/special\ record/file\ 3.mp4
```

### How do you do it today?

I make sure the `reolink-cli` is run from within the directory that files are to be downloaded into. A separate script runs after completion to tag files with their `recordType` from the JSON returned by `vod search`

### Camera models this applies to

all

reolink-net:
Added `--directory` for the next release — but I took only half of what you
proposed, so here is the reasoning.

**What landed:**

```
reolink-cli -c cam-i vod download 0120260825224024 0120260825223942 --directory /path/to/dir
```

Created if missing, works for one recording or many, each keeps its own name.
`--file` stays single-recording-only and the two are mutually exclusive.

**What I left out — the per-file naming.** Passing N filenames positionally
matched against N recordings is fragile in exactly the case that motivates it:
by the time you are downloading 300 recordings, keeping two lists in the same
order is worse than renaming afterwards. And the information is already there —
`vod download` prints a `file` field per recording, so a post-processing pass
has everything it needs. You mentioned you already run one to tag by
`recordType`, so this would be a second way to do something your pipeline
already does.

If you disagree, say so — the case I would want to hear is one where renaming
afterwards genuinely does not work, rather than being an extra step.

Two related things you may not know, both already true today:

- `--file` accepts a path, and creates the parent directories:
  `--file "/special record/file 3.mp4"` works for a single recording.
- Downloads are named `.mp4` from the next release (#94) — they were always an
  MP4 container, just mislabelled.

[...]

Issue #92, opened 2026-08-27
--password-stdin needs to be fixed when using reolink-cli device add


djancak:
### Version

reolink-cli 0.13.1 (external · LAN-only)

### Platform

Linux x86_64

### Camera model and firmware

any

### Command you ran

```shell
reolink-cli device add "front door" --host 192.168.1.41 --user admin --password-stdin
```

### What you expected, and what happened instead

Expected a password prompt, but got nothing at all. Had to CTRL+C out of the application.

It would be good to make `--password-stdin` function with any other areas of code where `--password` can be used as well

### Output

```shell
❯ reolink-cli device add "front door" --host 192.168.1.41 --user admin --password-stdin
nothing
is
happening
^C⏎

reolink-cli --gateway-addr 127.0.0.1:9000 --host 192.168.0.128 --user admin --password-stdin vod search --from 2026-08-25T08:00:00 --to 2026-08-25T12:00:00
error: unexpected argument '--password-stdin' found
```

reolink-net:
Both halves confirmed and fixed for the next release.

**The hang.** `--password-stdin` went straight to a blocking read of stdin, so
on a terminal it waits for EOF with nothing on screen — indistinguishable from a
hang, which is exactly what you hit. It now says what it is waiting for before
blocking (on stderr, so a piped password is unaffected and stdout stays clean
for the JSON):

```
reading the password from stdin — type it and press Ctrl-D (Ctrl-Z then Enter on Windows).
To be prompted instead, drop --password-stdin; to pipe it, use `echo <pw> | reolink-cli …`.
```

**The missing flag.** You are right that it should work wherever `--password`
does — and the irony was that commands which *didn't* have it printed
"use `--password-stdin`" in their error message, advice you could not follow.
It is a global option now, mutually exclusive with `--password`:

```
echo <pw> | reolink-cli --host 192.168.0.128 --user admin --password-stdin vod search --from … --to …
```

Verified against a real device: piped password logs in, a wrong one is rejected
(`authentication failed`), empty stdin gives `empty password read from stdin`
instead of hanging, and `device add --password-stdin` still works as before.

reolink-net:
Released in [v0.14.1](https://github.com/reolink/reolink-cli/releases/tag/v0.14.1).

`--password-stdin` is global now, so the form that failed for you works:

```
echo <pw> | reolink-cli --gateway-addr 127.0.0.1:9000 --host 192.168.0.128 \
[...]

Issue #91, opened 2026-08-26
Rules affecting device names


ian1182:
### Version

0.140.0

### Platform

Windows x86_64

### Camera model and firmware

n/a

### Command you ran

```shell
reolink-cli device add front door --host 192.168.1.41 --user admin --password wrong
```

### What you expected, and what happened instead

What are the rules defining acceptable device/camera name formats?

The password of "wrong" in my command is of course not valid :-)

When I run the above command via the windows command line the response is

```
warning: --password on the command line is visible in `ps` and shell history; prefer --password-stdin, the REOLINK_PASSWORD env var, or a registered camera (`device add`).
**error: unexpected argument 'door' found**

Usage: reolink-cli device add [OPTIONS] <ALIAS>

For more information, try '--help'.
```

If I however change the name to front-door or front_door its fine. This implies the name should contain no spaces and the placeholder entry uses front-door. Yet in the Install section under AI agents I see names with what appears to be spaces (e.g. “show me the front door camera”). When I add a camera/NVR/etc its possible to use spaces and other punctuation in the name which is accepted by the mobile app or desktop client but it seems not by the cli?

This is more a question of how names such as "front door" should be handled in the CLI.

### Output

```shell

```

djancak:
> If I however change the name to front-door or front_door its fine. This implies the name should contain no spaces and the placeholder entry uses front-door. Yet in the Install section under AI agents I see names with what appears to be spaces (e.g. “show me the front door camera”). When I add a camera/NVR/etc its possible to use spaces and other punctuation in the name which is accepted by the mobile app or desktop client but it seems not by the cli?

You can use "quotation marks" to create a name that has a space in it. Using quotation marks like this is the typical way that command-line arguments that contain a space in them are passed. The quotation marks themselves will not be part of the name.

`reolink-cli device add "front door" --host 192.168.1.41 --user admin --password wrong` should work as you expect

reolink-net:
@djancak is right — quoting is the answer, and it is a shell rule rather than a
reolink-cli one. `front door` reached the CLI as two arguments before the CLI
ever saw it, which is why the error came from argument parsing.

To answer the actual question — **there are no name rules**. The CLI does not
validate or normalise the alias at all, so anything your shell can pass through
works. I just checked each of these end to end on 0.14.0:

```
device add "front door"     ok
device add "front door 2"   ok
device add "Front Door"     ok
device add "客厅"            ok
device add "cam.1"          ok
```

[...]

PR #101, opened 2026-08-28
release: 0.14.1

cuican-reolink:
Release sync for **0.14.1**.

Lands `checksums/v0.14.1.sha256` **before** the release is published — the

### What is in 0.14.1

Everything here came from users of 0.14.0.

- **#94** — downloaded recordings were always named `.h264`. They are not an
  elementary stream at all: every file begins `…ftyp iso4`, an **MP4
  container**. The old sniffer looked for Annex-B start codes, found the
  `00 00 00 01` that happens to sit inside an MP4 header, read a meaningless
  NAL type, and fell back to `h264` — which is why *every* download got that
  name whatever the codec. Downloads are now `<recording>.mp4`.
- **#92** — `--password-stdin` blocked on a terminal with nothing on screen,
  which reads as a hang. It now says what it is waiting for, in wording checked
  under a pty: a single Ctrl-D after the password does not end the read, so the
  hint says *press Enter, then Ctrl-D*. The flag is also global now, matching
  `--password` — it previously existed only on `device add` / `device update`
  while every other command advised using it.
- **#93** — `vod download --directory <DIR>` chooses where recordings land,
  created if missing, for one recording or many. Previously the only way was to
  `cd` first, which an MCP client or an agent cannot do.
- **#91** — camera names have no rules (spaces, mixed case, non-ASCII all
  work), but a name with a space needs shell quotes and every example showed
  the hyphenated form. Documented in `device add --help`, the command reference
  and the bundled skill.

### Verification

  `clean (external)` for every one. The finished archives were scanned too: 0
  fingerprints and 0 bundled P2P libraries, against 170 in an internal build
  used as a positive control, so the scan is known to work.
- `reolink-cli --version` → `0.14.1 (external · LAN-only)`, and the shipped
  binary carries the new `--directory` flag.
  per-artifact sidecars, 8/8.
- Verified against a real Reolink Home Hub 2 (v3.3.0.579): downloads land as
  `.mp4` and `ffprobe` agrees, `--directory` creates and fills the directory,
  and a piped password logs in while a wrong one is rejected.
- The terminal behaviour of `--password-stdin` was verified under a pty, which
  is how the first version of the hint was caught being wrong.
- 749 tests pass. Two of the new ones exist to stop a flag pair silently doing
  the wrong thing: `--file` with `--directory`, and `--password` with
  `--password-stdin`, both must refuse rather than quietly ignore one.

Installer end-to-end runs (macOS + Alpine on both architectures) come after the
release is published, since they resolve the latest release.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
     {
       "name": "reolink-cli",
       "description": "Operate Reolink cameras locally via reolink-cli: discover, login, inspect, configure, preview, snapshot, and control PTZ/light/audio/detection/recording.",
-      "version": "0.14.0",
+      "version": "0.14.1",
       "source": "./",
       "author": {
         "name": "Reolink"
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "reolink-cli",
   "description": "Operate Reolink cameras locally: discover, login, inspect, configure, preview, snapshot, and control PTZ/light/audio/detection/recording.",
-  "version": "0.14.0",
+  "version": "0.14.1",
   "author": {
     "name": "reolink"
   },
--- a/.codex-plugin/plugin.json
+++ b/.codex-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "reolink-cli",
-  "version": "0.14.0",
+  "version": "0.14.1",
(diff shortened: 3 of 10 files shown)


reolink-net (approved):
Checksum lands before the release. All eight external artifacts built with the P2P gate clean and the finished archives scanned against a positive control; every fix verified on a real Home Hub 2, and the --password-stdin terminal wording checked under a pty.

Issue #99, opened 2026-08-28
Expose child-camera identity and per-channel recording access behind Home Hub


ridome:
### What are you trying to do?

Treat each camera connected to Home Hub as an addressable child resource. For a selected Hub channel, the CLI should expose the child camera's model, firmware, hardware version, name, and per-channel capabilities, and then use that same identity when searching and downloading the child's recordings.

The output should clearly separate Hub-level metadata from child-camera metadata so agents can make capability-aware decisions without assuming that the Hub model represents every channel.

### How do you do it today?

When trying to view recordings for a child device, the CLI can only retrieve information for the Home Hub itself. It cannot query the individual IPC model on each channel. As a result, the agent cannot reliably identify the child device, determine its capabilities, or complete a channel-specific recording workflow.

### Camera models this applies to

Home Hub with one or more IPC child cameras; exact models and firmware versions were not captured.

reolink-net:
Implemented for the next release.

The cause: `info` sends cmd 80 `GetVersion`, which answers for the **device**.
On a hub every channel therefore reported the hub, and the child cameras were
invisible — exactly as you describe:

```
hub-ch0  model=Reolink Home Hub 2
hub-ch1  model=Reolink Home Hub 2
hub-ch2  model=Reolink Home Hub 2
```

There is a per-channel equivalent, cmd 318 `NET_GET_CHN_VERSION_V20`, which was
simply not implemented. `info` now sends both and reports the child under its
own `channel` object:

```
hub-ch0  hub=Reolink Home Hub 2 | child: ch0 Argus PT Ultra   / v3.0.0.4739_25040737 / Front Door
hub-ch1  hub=Reolink Home Hub 2 | child: ch1 Reolink Argus PT / v3.0.0.5585_25102990 / Back Door
hub-ch2  hub=Reolink Home Hub 2 | child: ch2 OMVI 2i Ultra    / v3.0.0.6976_26081100 / Back Yard
```

Kept as a separate object rather than merged into the top level, for the reason
you gave: a caller deciding what a camera supports must not mistake hub
metadata for the child's. Serial number, hardware version and build day come
through as well.

On the recording half of your request — searching and downloading a child's
recordings by channel already works. Verified on the same hub:
`vod search --from … --to …` against `hub-ch0` returned that channel's
recordings, and downloading one produced a 3.0 MB file. If you saw otherwise it
was likely one of the two download bugs fixed in
[...]

Issue #97, opened 2026-08-28
Support manual spotlight control for cameras connected through Home Hub


ridome:
### What are you trying to do?

Allow a Home Hub child camera to be targeted by its registered name or Hub channel and manually turn its visible white spotlight on or off. The CLI should distinguish manual spotlight control from alarm-triggered white-LED configuration, discover the capability per child channel, apply the command to the child camera rather than the Hub, and read back the resulting state.

### How do you do it today?

Manual white-light control works as a user intent, but a camera connected underneath Home Hub cannot currently be made to turn on its white light through the CLI. The device is reachable through the Hub, yet the light command is not routed or exposed for the child channel.

### Camera models this applies to

A spotlight-equipped Front Door camera connected through Home Hub; exact child-camera model, Hub model, and firmware versions were not captured.

reolink-net:
Partly a real defect, partly a wrong premise — and the honest answer needs both
halves.

**The premise.** Spotlight control does reach Home Hub child cameras today. On
a Home Hub 2 with three child cameras, all three turn their white light on and
off through `light spotlight set --enable|--disable` addressed at the child
channel. So it is routed and exposed.

**The defect.** The first command to a channel sometimes comes back `400`, and
everything after it works:

```
hub-ch0  three calls: 400  ok  ok
hub-ch1  three calls: ok   ok  ok
```

Try it once and you conclude the feature is unsupported, which is what
happened here.

I could not pin the trigger, and I would rather say so than invent one. It is
not idle time — after letting the camera sit for 120s the first call
*succeeded*. It is not a cold gateway either — restarting the gateway and
sending the spotlight command first produced `ok` on all three channels. It
does not reproduce on demand.

So the fix does not pretend to know why: a bounded backoff retry (400 ms,
1200 ms, 2500 ms) on a device-side `400`, with `retried` in the response so you
can see it happened. A `405` — the model genuinely has no spotlight — is not
retried, since that only spends seconds arriving at the same answer.

Verified on device: a channel that answered `400` now returns
`{"state":"on","retried":true}` after ~6.8 s, channels that answer immediately
return in ~0.3 s with `retried: false`, and a genuinely unsupported command
still fails in under 100 ms.

Also landing in the same release: `info` on a hub channel now reports the child
camera's own model and firmware (#99), which is what you need to know whether a
given child has a spotlight at all.

reolink-net:
[...]

Issue #95, opened 2026-08-28
Support verified, duration-bounded siren activation


ridome:
### What are you trying to do?

When an agent receives a request such as "sound the Front Door siren for 5 seconds", the CLI should provide a deterministic operation that:

- starts the siren;
- stops it after the requested duration;
- verifies that the siren is off afterward; and
- returns machine-readable evidence such as the requested duration, actual start/stop timestamps, and final state.

If the camera firmware owns the timeout, the CLI should read back the final state. If the CLI owns the timeout, it should guarantee cleanup on interruption or timeout and report a failure instead of an unverified success.

### How do you do it today?

In an agent test, the response was:

> Manually triggered the Front Door siren for 5 seconds (state: on; it will stop automatically after a few seconds).

This confirms only that the command returned state: on. It does not verify that the siren ran for exactly five seconds or that the final state returned to off, so an agent can report success without end-state evidence.

### Camera models this applies to

Front Door camera; exact model and firmware were not captured. This should also work for supported cameras connected through Home Hub.

reolink-net:
You are right, and the cause is worse than an imprecise message: `state: "on"`
was a hard-coded string. The CLI wrote it whenever the device acknowledged the
command, so it could never have been evidence of anything.

Reproduced on a Home Hub 2. Watching the event stream while firing the siren
three times at a child camera:

```
cmd 263 → 200 on two channels, 400 on a third
event stream → no siren.on at all (only an idle "none")
```

A device that really sounds pushes cmd 547, which the event bus surfaces as
`siren.on` / `siren.off`. Nothing arrived. So the agent's "state: on" was
faithfully repeating something the CLI made up.

Fixed for the next release. v2.0 has no GET for siren state, so rather than
invent verification the response now reports only what is actually known:

```json
{
  "accepted": true,
  "requestedDuration": 5,
  "sentAt": 1787883774,
  "expectedOffAt": 1787883779,
  "verified": false,
  "note": "the device accepted the command; it does not report siren state on request. Confirm with `events stream` — a device that sounds emits siren.on / siren.off."
}
```

`verified: false` is deliberate and permanent for this transport: an agent
reading this can no longer claim the siren sounded, only that the command was
accepted. Where the device does push `siren.on`, `events stream` is the
confirmation channel — which is the part of your request the protocol can
actually satisfy today.

Leaving this open: you also asked for guaranteed cleanup on interruption and a
failure instead of an unverified success. The second half is done; the first
[...]

PR #102, opened 2026-08-28
release: 0.14.2

cuican-reolink:
Release sync for **0.14.2**.

Lands `checksums/v0.14.2.sha256` **before** the release is published — the
installers resolve the latest release and then look for its checksum file here,
and they fail closed. The other order breaks `curl | sh` for everyone in
between.

### What is in 0.14.2

Three Home Hub defects, all reported by @ridome and all verified on a Home Hub 2
with three child cameras (Argus PT Ultra, Reolink Argus PT, OMVI 2i Ultra).

- **#99** — a hub channel reported the hub's own model. `info` sends a
  device-level command, so every channel answered `Reolink Home Hub 2` and the
  children were invisible. The per-channel equivalent existed in the protocol
  and was simply not implemented; the child now appears under its own `channel`
  object, kept separate so hub metadata is never mistaken for a child's
  capabilities.
- **#95** — `audio siren play` claimed the siren was sounding on nothing more
  than a device acknowledgement. `state: "on"` was hard-coded. Measured: the
  command returned 200 while no `siren.on` report arrived, so an agent relayed
  a success the CLI had invented. The response now reports `accepted`,
  timestamps and `verified: false`, and points at `events stream` for the
  confirmation the protocol can actually provide.
- **#97** — the first spotlight command to a hub child could fail `400`, making
  a working feature look unsupported. The trigger could not be pinned (it did
  not correlate with idle time or a fresh gateway), so the fix retries on `400`
  with a bounded backoff and reports `retried` rather than inventing a cause. A
  `405` — no spotlight on that model — is not retried.

### Verification

- All eight external artifacts built; the P2P gate reports `clean (external)`
  for every one. The finished archives were scanned too: 0 fingerprints and 0
  bundled P2P libraries, against 170 in an internal build used as a positive
  control.
- `reolink-cli --version` → `0.14.2 (external · LAN-only)`.
- Checksums recomputed independently and matched against the sidecars, 8/8.
- On device: three channels return three distinct child models; the siren
  response contains no unverified `state: "on"`; a channel that answered `400`
  succeeded after retry in 6.8s while a healthy one returned in 0.3s and a
  genuinely unsupported command still failed in under 100ms.
- 752 tests pass, including three new ones asserting that only a `400` is
  retried — a `405` or an auth/transport failure must not be.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
     {
       "name": "reolink-cli",
       "description": "Operate Reolink cameras locally via reolink-cli: discover, login, inspect, configure, preview, snapshot, and control PTZ/light/audio/detection/recording.",
-      "version": "0.14.1",
+      "version": "0.14.2",
       "source": "./",
       "author": {
         "name": "Reolink"
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "reolink-cli",
   "description": "Operate Reolink cameras locally: discover, login, inspect, configure, preview, snapshot, and control PTZ/light/audio/detection/recording.",
-  "version": "0.14.1",
+  "version": "0.14.2",
   "author": {
     "name": "reolink"
   },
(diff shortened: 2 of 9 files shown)


reolink-net (approved):
Checksum lands before the release. Eight external artifacts with the P2P gate clean and the archives scanned against a positive control; all three fixes verified on a real Home Hub 2.

Issue #100, opened 2026-08-28
Support both views of dual-lens cameras connected through Home Hub


ridome:
### What are you trying to do?

Discover and expose both physical views of a dual-lens camera connected through Home Hub. Each view should have a stable machine-readable identifier and expose its own preview, snapshot, stream, recording, and capability information where supported.

Main/sub stream quality profiles must remain separate from physical lens or view selection. An agent should be able to request the fixed-lens view or the PTZ view explicitly instead of receiving only one channel with two encoding profiles.

### How do you do it today?

For the tested bullet-plus-PTZ child camera, the CLI reported one Back Yard channel with these profiles:

- Main: 3840x2160, H.265, 15 fps, 2048 kbps
- Sub: 896x512, H.264, 15 fps, 672 kbps

These are main/sub encodes for one exposed view, not the two physical views of the device. There is no CLI selector or inventory entry for the secondary lens when the camera is connected through Home Hub.

### Camera models this applies to

A dual-lens bullet-plus-PTZ camera connected through Home Hub; exact product and firmware versions were not captured.

reolink-net:
Investigated on a Home Hub 2 with exactly this kind of child — a bullet+PTZ on
channel 2. **The device does report both views; the CLI does not expose them.**

Two pieces of evidence.

`GetSupport` (cmd 199) returns a `<subItem>` block per sub-view, and only the
dual-lens channel has more than one:

```
chnID=0  subItems=0
chnID=1  subItems=0
chnID=2  subItems=2     ← the bullet+PTZ
chnID=3..7  subItems=0  (empty slots)
```

The two differ in exactly the way the two physical views should:

| sub-view | chnID | ptzType | ptzPreset | reading |
|---|---|---|---|---|
| subItem[0] | 0 | — | 0 | fixed lens — no PTZ presets |
| subItem[1] | 1 | 3 | 1 | PTZ lens — has a PTZ type and presets |

And cmd 738 `NET_GET_CHANNEL_LENSTYPE_INFO_V20` answers on that channel alone:

```
ch0 → 400   ch1 → 400   ch2 → <lensTypeInfo><lensType>wide</lensType></lensTypeInfo>
```

`lensType` is a single read-only value (`wide` / `telephoto` per the SDK
header), so it labels a view rather than enumerating them — the enumeration is
the `subItem` list above.

So your reading is right: the Main/Sub you saw are encoding profiles of one
exposed view, and the second view is addressable in the protocol but has no CLI
[...]

Issue #96, opened 2026-08-28
Add TTL-scoped event monitoring suitable for agent tasks


ridome:
### What are you trying to do?

Allow an agent to create a bounded monitoring task such as: monitor Front Door channel 0 for people for the next 30 minutes; capture a snapshot and a short clip for each event; make the captured media available to a downstream classifier; notify only when the classifier identifies a delivery; and expire the task automatically after 30 minutes.

The monitoring job must continue independently of whether the chat remains active. Its status, expiry time, captured events, and final result should be queryable as structured data. Temporary rules should clean themselves up automatically.

### How do you do it today?

The current agent workflow can start the event-monitor daemon and add a persistent rule that writes snapshots to an outbox. However, the agent reports that it cannot wake itself while the conversation is inactive. The user must return later, ask for a review, and manually clean up the temporary rule.

The daemon provides durable capture, but the CLI does not expose the complete bounded-task lifecycle needed for a request such as "watch this camera for the next 30 minutes and only tell me if it is a delivery person."

### Camera models this applies to

Front Door camera on channel 0; exact model and firmware were not captured.

reolink-net:
Not taking this one on yet, and I would rather say why than leave it silent.

What exists today covers the durable half: `events monitor` is a
SIGHUP-reloadable daemon that matches events against TOML rules and runs actions
(snapshot, audio talk, exec hook, log), independently of any chat session. What
you are asking for on top of it is a **bounded task lifecycle**: create with a
TTL, query status and captured events as structured data, expire and clean up
the temporary rule automatically, and hand the captured media to a classifier.

That is a real gap, but it is a feature with genuine design questions rather
than a fix:

- where task state lives, and what happens to a running task when the daemon
  restarts;
- whether expiry removes the rule or just deactivates it, and what happens to
  media already captured;
- what "make the media available to a downstream classifier" means concretely —
  an outbox directory the agent polls, an exec hook, or something the CLI
  invokes;
- and whether notification is the CLI's job at all, or the agent's once it has
  queryable results.

Guessing at those and shipping would be worse than waiting: a task lifecycle is
hard to change once scripts depend on it.

The honest position on the agent limitation you hit: no CLI change can make an
agent wake itself. What the CLI can do is make the daemon's captures queryable
after the fact so the agent's "come back and review" step is reliable and the
cleanup is automatic — that is the part worth building.

If you want to push this forward, the most useful thing would be your answers to
[...]

PR #103, opened 2026-09-01
release: 0.15.0

cuican-reolink:
Release sync for 0.15.0. Binaries are built and verified; this lands the
checksum file on `main` first so `curl | sh` never resolves a release whose
hash is missing.

## What is in it

**The siren was accepted and silent in every release before this one** (#95).
0.14.2 stopped the CLI *claiming* the siren had sounded — correct, but it never
established whether it sounded at all. It did not. Cmd 263 carries a
`playMode`; the CLI sent mode 1 ("sound for N seconds"), which this firmware
accepts with a `200` and ignores. Mode 0 ("sound N times") works. Measured by
capturing the camera's own microphone through the preview stream and firing
both modes into one recording — flat at ambient on mode 1, roughly forty times
higher on mode 0.

**A `400` from a sleeping camera read as "your model does not support this"**
(#95, #97). A battery child behind a hub sleeps after about two minutes; a
sleeping child rejects commands with the same code the device uses for a
command it does not implement. That is how an agent came to tell a user their
Argus PT Ultra had no siren. Bounded retry, `attempts` in the answer, and an
error that explains standby instead of passing "unsupported" through.

**`vod search` failed for a day at the start of every month.** A window
crossing a month end produced no query windows internally, so the search read
as "no recordings"; a guard turned that into an outright rejection. The window
is now walked by calendar date.

Also: `--view N` for dual-lens cameras (#100), bounded monitoring tasks via
rule `expires_at` and `history --rule/--since` (#96), per-file skip in batch
`vod download`, `speakerVolume` in the siren answer, and the two Wi-Fi/gateway
timeout fixes built internally as 0.14.3 and never released on their own.

Full notes in `CHANGELOG.md`.

## Release safety

- All eight external artifacts built with `EXTERNAL=1`; `strings` finds **zero**
  P2P fingerprints in any of them. Positive control: the internal 0.14.2
  tarball scans **80**, so the scanner is doing something.
- `reolink-cli --version` on the packaged macOS binary reports
  `0.15.0 (external · LAN-only)`.
- `checksums/v0.15.0.sha256` is copied from the build machine's
  `dist/SHA256SUMS`, not re-downloaded from a release.
- `./scripts/check-version-sync.sh 0.15.0` passes; the
  "committed hashes vs release" row skips because the release is not published
  yet, which is the intended order.
- `./scripts/check-docs-mirror.sh` was run before the copy: every line the sync
  would overwrite is an older revision of a line changed here, so no
  public-only content is lost.

The release itself is published after this merges.


reolink-net (approved):
Verified before approving: all eight external artifacts scan clean for P2P fingerprints (positive control on an internal build scans 80, so the check is doing something), the packaged binary reports `external · LAN-only`, and `checksums/v0.15.0.sha256` comes from the build machine's SHA256SUMS rather than a re-download. `check-version-sync.sh 0.15.0` passes with the release row correctly skipped — the checksum lands before the release, not after.

PR #104, opened 2026-09-02
release: 0.16.0

cuican-reolink:
Release sync for 0.16.0.

Two snapshot bugs, both ours. They were found by timing each phase of a capture
(connect / login / device-produces-image / transfer / logout) instead of
assuming the wait was the camera's physics — the phase that dominated turned
out to be one we were causing.

### `<fullFrame>` was overriding `--stream`

It is a model-specific mode (the reference SDK documents it as "snap full frame
pic, *for sd7*", to be used *together with* the sub stream). The CLI sent
`fullFrame=1` to every device unconditionally. On a camera that implements the
flag that overrides `<streamType>`, so asking for the small preview JPEG got you
a full-resolution one.

Measured on a dual-lens child behind a Home Hub 2, same camera, otherwise
identical request:

```
                  resolution     size     awake     asleep (first capture)
--stream sub
  fullFrame=1     3840x2160     875 KB     9.34 s     23 s
  fullFrame=0      896x512       58 KB     1.92 s     6.46 s

--stream main
  fullFrame=1     3840x2160     861 KB     17.1 s
  fullFrame=0     3840x2160     860 KB      5.2 s
```

`--stream main` is worth a second look: same resolution, same bytes to within
0.1%, so the image itself is unchanged — but the flag was routing the capture
down a much slower and far less stable path (three runs: 11.8 / 17.1 / 28.8 s
with it, 4.7 / 5.2 / 7.4 s without). Cameras that never implemented the flag
measured the same either way, which is why only one model ever looked broken —
and it looked like it was ignoring `--stream` rather than obeying an
instruction we should not have been sending.

### A 10 s read timeout turned slow captures into failed ones

The transport budget bounds a *single* read, and a snapshot's cmd 109 is one
long silence: a sleeping battery camera brings its whole video pipeline up
before sending the first byte. Measured 5.1 s for a sub-stream capture from a
sleeping child, and 23 s for the 4K child with three channels waking at once —
so anything past 10 s came back as `tcp read timeout after 10s` instead of an
image. Raised to 30 s, which covers the wake and stays inside the CLI's own
60 s response budget. Kept as one global value rather than a per-command
argument: VOD download and preview first-frame have the same single-long-wait
shape.

### Added

`events history` — search the event log the device itself recorded (hub / NVR,
cmds 516/517/518). Separate store from `events query`, which reads the
gateway's in-memory ring of live pushes (recent ~500, lost on restart).

### Release artifacts

All 8 platforms built external (LAN-only, no P2P); the P2P fingerprint gate
reports `clean (external)` on every one. `checksums/v0.16.0.sha256` is copied
from the build machine's `dist/SHA256SUMS`, not re-downloaded.

Verified end to end with the packaged binaries themselves (both `reolink-cli`
and `reolink-gateway` from the artifact, since the capture path lives in the
gateway): `--stream sub` returns 896x512 in 2.42 s, `--stream main` returns
3840x2160 in 5.34 s.


reolink-net (approved):
Release sync verified: 8 external artifacts built with the P2P gate clean on each, checksums copied from the build machine, and the packaged binaries tested end to end against a real hub.

PR #106, opened 2026-09-04
release: 0.16.1

cuican-reolink:
Three fixes, all cases where the CLI knew what had gone wrong and reported something else.

- A failed single-file `vod download` reported the hint reserved for an empty stream — advice about battery cameras behind a hub — instead of the device's error, because the reader stopped at the skip marker that precedes it (#105). Reproduced on a Home Hub with a name that does not exist: one name gave the battery text, two names (the batch path) gave `device rejected command 14 (400)`.
- "Recording not found" now names the channel it searched, and the channel the recording name points at.
- Aliases carrying both a `host` and a `uid` could not use `snapshot`, `preview` or `vod download` at all: login routed by uid, the media URL named the IP, and the gateway rejected its own token with `token does not match host`. Not a regression.
- Preview no longer dies ~11s in on battery devices; the heartbeat answer lived in only one of the two receive loops.

Artifacts built external (LAN-only, no P2P) for all 8 platforms; the P2P fingerprint gate reports clean on every one, and `checksums/v0.16.1.sha256` comes from the build machine.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
     {
       "name": "reolink-cli",
       "description": "Operate Reolink cameras locally via reolink-cli: discover, login, inspect, configure, preview, snapshot, and control PTZ/light/audio/detection/recording.",
-      "version": "0.16.0",
+      "version": "0.16.1",
       "source": "./",
       "author": {
         "name": "Reolink"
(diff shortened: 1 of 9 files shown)


reolink-net (approved):
Release sync verified: 8 external artifacts, P2P gate clean on all, checksums from the build machine.

PR #107, opened 2026-09-04
release: 0.17.0

cuican-reolink:
`vod download` can now ask the camera for a time range and let it do the cutting (#105).

```sh
reolink-cli vod download --from 2026-09-02T09:45:00 --to 2026-09-02T09:55:00 --channel 5 --directory ./clips
```

There is no protocol document for the device command behind this, so the behaviour was measured rather than quoted:

- **One request returns one recording** — the first the window touches, clipped to it; a window starting in a gap skips to the next. Verified by frame count on four windows (241 / 749 / 749 / 781 frames). So the CLI searches the window and asks once per recording, joining the pieces — which is what makes a window crossing the hour come back whole (three recordings → 2165 frames, 8.87 MB, 3.4 s).
- **The output is an elementary stream** (`.hevc`/`.h264`), not MP4: a cut is produced live rather than read from a stored file.
- **Not every model implements it**, and nothing in the capability list advertises it. Those answer 400; download-by-name is unaffected.

Also fixed: `--directory` used a file name guessed before the request went out, so a raw HEVC cut landed in a file called `.mp4`.

Artifacts built external (LAN-only, no P2P) for all 8 platforms; the P2P fingerprint gate reports clean on every one, and `checksums/v0.17.0.sha256` comes from the build machine.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
     {
       "name": "reolink-cli",
       "description": "Operate Reolink cameras locally via reolink-cli: discover, login, inspect, configure, preview, snapshot, and control PTZ/light/audio/detection/recording.",
-      "version": "0.16.1",
+      "version": "0.17.0",
       "source": "./",
       "author": {
         "name": "Reolink"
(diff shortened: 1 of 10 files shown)


reolink-net (approved):
Release sync verified: 8 external artifacts, P2P gate clean on all, checksums from the build machine; feature verified on hardware (frame counts on four windows plus a three-recording join).

PR #110, opened 2026-09-06
release: 0.18.1

cuican-reolink:
Both of these were found only after the reporter of #105 ran the real thing and posted what came back.

**The by-name download's probe is rejected by an NVR (400) — the original problem in #105.** On 0.18.0 his download answered `device rejected command 13 (400)`: the *first* step of the two-step handshake, before any data. That probe had been cut down to a bare `<name>` because a Video Doorbell SE rejected one carrying `streamType` *and* `channelId` — but the reference implementation's own probe carries `supportSub` + `streamType` and no `channelId`, so only `channelId` deserved to go and the other two were collateral. The probe now matches the reference field for field; downloads by name were re-tested on both streams.

**A time-range download now states its frame rate — without it `-c copy` produced a clip of the wrong length.** An elementary stream carries no timestamps, so ffmpeg invents a rate: every frame present, duration wrong.

The note in 0.18.0 claiming the remux measured 49.856 s "within one frame" was wrong — that read the container duration, which reports the longest track (the audio). Per track the video was **2.0007 s**: 749 frames packed into two seconds. The device states the rate in the stream's opening packet and we were discarding it; it now travels as `X-Frame-Rate`, appears as `frameRate` in the CLI output, and the CLI prints the command to copy:

```
ffmpeg -r 15 -i clip.hevc -i clip.aac -c copy clip.mp4
```

`-r` goes **before** `-i` — it sets the *input* rate. Measured per track after the fix: video 49.933 s at 15 fps over 749 frames, audio 49.856 s, against 49.828 s for the same recording downloaded by name. The 105 ms gap is the device running at 15.02 fps while we state a whole 15 — 0.2%, nothing re-encoded.

Artifacts built external (LAN-only, no P2P) for all 8 platforms; the P2P fingerprint gate reports clean on every one, and `checksums/v0.18.1.sha256` comes from the build machine.


reolink-net (approved):
Release sync verified: 8 external artifacts, P2P gate clean on all, checksums from the build machine. Probe change re-tested on hardware for both streams; frame rate verified per track after remux.

PR #111, opened 2026-09-07
release: 0.18.2

cuican-reolink:
**A cut's reported frame rate came from the wrong place, and remuxing against it changed the clip's length (#105).**

0.18.1 was right to pass a rate through and wrong about where to read it: it used the byte in the stream's own info header, which is correct on the sub stream and wrong on the main one, where it announces a nominal 30 rather than what is delivered.

Measured on a Home Hub 2, main stream: the encoder reports 15, an 11-second cut holds 162 frames (14.7 fps), and the header claims 30. Remuxed at 30, a 10.8-second clip became 5.4 — exactly half. The reporter's 25 fps cameras got the same 30, turning an 11:00 clip into 9:10; 25/30 is precisely that ratio.

The rate now comes from the channel's encoder configuration, the source that matched the frame count on every measurement. A device that does not answer that command simply gets no `-r` hint and the cut itself is unaffected. A test pins the header out of the path.

Re-verified before release on a second camera: the encoder reports 15 for the main stream, the cut reports 15, and the remuxed clip measures 49.933 s over 749 frames against 49.828 s for the same recording downloaded whole by name.

Artifacts built external (LAN-only, no P2P) for all 8 platforms; the P2P fingerprint gate reports clean on every one, and `checksums/v0.18.2.sha256` comes from the build machine.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
     {
       "name": "reolink-cli",
       "description": "Operate Reolink cameras locally via reolink-cli: discover, login, inspect, configure, preview, snapshot, and control PTZ/light/audio/detection/recording.",
-      "version": "0.18.1",
+      "version": "0.18.2",
       "source": "./",
       "author": {
         "name": "Reolink"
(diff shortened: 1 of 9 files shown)


reolink-net (approved):
Release sync verified: 8 external artifacts, P2P gate clean on all, checksums from the build machine. Frame-rate source re-verified on hardware before release.

PR #112, opened 2026-09-07
release: 0.18.3

cuican-reolink:
Found while checking a report that `discover` does not see battery cameras. It does not, and there are two separate reasons.

**A Home Hub with a single paired camera can be expanded now.** `device expand` answered this, on a device that *is* a Hub and has a battery video doorbell on channel 0:

```
camera hub has a single populated channel (0); nothing to expand (not an NVR or Hub)
```

— an inference stated as fact, and backwards. It treated "one populated channel" as "this is a plain camera", but one camera and one channel are different claims, and a hub with a single paired camera is the case where expanding matters most: that child has no address on the network, so `discover` cannot see it and this command is the only way to register it. The device already says which it is — the channel count reads 8 on that hub and 1 on a plain camera — so the refusal now applies only when the device itself does not claim more than one channel. Verified: expanding registers the child, and querying that entry reaches the doorbell.

**`device expand`'s help said Home Hub was unsupported.** Only `protocol=v30` entries are refused; a Home Hub registered as v20 has always worked.

**`discover` now says what it cannot find**, in its own `--help`: a sleeping battery camera (the probe is answered by the camera's main application, powered down while it sleeps — the same camera awake answers every time) and a camera paired to a Hub or NVR (no address on the network at all; the parent answers). Raising the timeout does not help — 2 s and 8 s return the same list.

Artifacts built external (LAN-only, no P2P) for all 8 platforms; the P2P fingerprint gate reports clean on every one, and `checksums/v0.18.3.sha256` comes from the build machine.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
     {
       "name": "reolink-cli",
       "description": "Operate Reolink cameras locally via reolink-cli: discover, login, inspect, configure, preview, snapshot, and control PTZ/light/audio/detection/recording.",
-      "version": "0.18.2",
+      "version": "0.18.3",
       "source": "./",
       "author": {
         "name": "Reolink"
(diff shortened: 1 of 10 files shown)


reolink-net (approved):
Release sync verified: 8 external artifacts, P2P gate clean on all, checksums from the build machine. Expand fix verified on a Home Hub with a battery doorbell child.

PR #113, opened 2026-09-09
release: 0.19.0

cuican-reolink:
**Breaking: a time-range download (`vod download --from/--to`) now produces MPEG-TS (`.ts`), and you must remove `-r <fps>` from the remux.** The timing now lives in the file; a rate on the command line overrides it and puts the old drift back. Downloading by name is untouched — that is still the MP4 the device stored.

```sh
reolink-cli vod download --from 2026-09-02T09:45:00 --to 2026-09-02T09:55:00 --channel 5 --file clip.ts
ffmpeg -i clip.ts -c copy clip.mp4                    # no -r
ffmpeg -i clip.ts -i clip.aac -c copy clip.mp4        # with --audio
```

**Why: the camera's timeline is not constant-rate, and no single frame rate can reconstruct it.** A cut clip drifted 0.2% against real time — 9 frames over 3 minutes, reported by someone cutting to an Ableton track in Premiere (#105). Two earlier attempts fixed *where the frame rate came from* (0.18.1 the stream's header byte, 0.18.2 the encoder's configuration); the real answer is that a scalar cannot describe this stream. Measured on a 40.6-second recording: **386 intervals of 80 ms and 198 of 40 ms**, and the device's own stored MP4 reports `avg_frame_rate=305500/20327`. Even remuxing at the exact average misplaces individual frames by up to 68 ms.

The device was sending the real timeline all along and we were discarding it: every video packet's BcMedia header carries that frame's timestamp in microseconds. It now travels with the video, in a container that can hold it.

Verified against the device's own file: a 30-second window came back as 451 frames whose 450 inter-frame gaps appear **verbatim** inside the gap sequence of the same recording downloaded whole by name. Remuxed, that clip measures 30.032 s at `avg_frame_rate=56375/3754`; the old path at a flat 15 fps gave 30.067 s. Both codec paths checked on real hardware (HEVC 3840×2160 main, H.264 896×512 sub).

**Also fixed**

- `ffmpeg -i clip.hevc -c copy` without `-r` produced a 2-second, 8183 fps file. An elementary stream states no timing, so forgetting the flag was a silent trap; a container removes it.
- A window spanning two recordings no longer plays the gap between them as a frozen frame — a window crossing a 9.5-minute gap now comes back as 800 frames over 53.1 s, largest inter-frame gap 81 ms.
- `discover`'s P2P leg was cut short by the LAN probe's timeout, so cameras only it can see appeared and disappeared between scans. The caller's timeout is now a floor: the scan ends after 3 quiet seconds, capped at 20. Before, the P2P leg returned nothing on 5 of 5 runs at default settings; after, 5 of 5 in 3.3 s.
- A UID-only target could fail immediately with `-7` right after a scan had just enumerated it — the LAN entry had not landed in the library's table yet. It now re-scans and retries, once, for at most 6 seconds. A UID already in the table pays nothing.

Artifacts built external (LAN-only, no P2P) for all 8 platforms; the P2P fingerprint gate reports clean on every one, and `checksums/v0.19.0.sha256` comes from the build machine.

### Install

```sh
curl -fsSL https://raw.githubusercontent.com/reolink/reolink-cli/main/install.sh | sh
```

All artifacts are **external** builds — LAN-only, no P2P. Verify before extracting:

```sh
shasum -a 256 -c reolink-cli-0.19.0-external-<platform>.<ext>.sha256
```

Checksums are also committed to `checksums/v0.19.0.sha256` on `main`, which is what the installers verify against.


reolink-net (approved):
Verified: 8 external artifacts, P2P fingerprint gate clean on all, checksums from the build machine.

